PR #65, opened 2022-01-26
Update api client, add support for big_number display type

smithclay:
* Adds basic rate limiting/retry logic to the API client to avoid issues with too many requests being sent to the API that results in 500s.
* Adds user-agent and organization as request headers for future analytics on provider usage.
* Adds support for `big_number` metric charts.

version `1.51.2`


client/client.go
```diff
@@ -44,8 +52,11 @@ type Body struct {
 type Client struct {
 	apiKey      string
 	baseURL     string
-	client      *http.Client
+	orgName     string
+	client      *retryablehttp.Client
+	rateLimiter *rate.Limiter
 	contentType string
+	context     context.Context
```
DanTulovsky, line 59: in general it's not correct to put contexts inside structs.  https://go.dev/blog/context-and-structs

Are you doing this for a particular reason?

client/client.go
```diff
@@ -60,31 +71,54 @@ func NewClient(ctx context.Context, apiKey string, orgName string, env string) *
...
+		c.context,
+		c,
 		fmt.Sprintf("%v/%v", c.baseURL, suffix),
 		httpMethod,
 		Headers{
-			"Authorization": fmt.Sprintf("bearer %v", c.apiKey),
-			"Content-Type":  c.contentType,
-			"Accept":        c.contentType,
+			"Authorization":   fmt.Sprintf("bearer %v", c.apiKey),
+			"User-Agent":      "terraform-provider-lightstep",
```
ishg, line 106: For future proofing and more analytics, might want to include the provider version too based on the User-Agent [standards](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)

client/client.go
smithclay: I am not, will remove :)

client/client.go
smithclay: Now with version + terraform sdk version

client/client.go
```diff
@@ -60,31 +77,54 @@ func NewClient(ctx context.Context, apiKey string, orgName string, env string) *
...
+// checkHTTPRetry inspects HTTP errors from the Lightstep API for known transient errors
+func checkHTTPRetry(_ context.Context, resp *http.Response, err error) (bool, error) {
+	if resp.StatusCode == http.StatusInternalServerError {
+		return true, nil
+	}
+	return false, nil
+}
+
 // CallAPI calls the given API and unmarshals the result to into result.
 func (c *Client) CallAPI(httpMethod string, suffix string, data interface{}, result interface{}) error {
```
DanTulovsky, line 104: I think the correct thing to do here is to pass in the context as first parameter...

DanTulovsky (changes requested):

DanTulovsky (approved):

client/client.go
smithclay: Just did a big pass on cleaning up context, now passed to each API call method

DanTulovsky (approved):

PR #68, opened 2022-02-02
LS-27435: Prevent exception when response is nil

smithclay:
* Fixes fatal exception when response object is nil in http client retry logic. (cc @gdvalle)

```
Stack trace from the terraform-provider-lightstep_v1.51.3 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xb236e5]

goroutine 114 [running]:
github.com/lightstep/terraform-provider-lightstep/client.checkHTTPRetry(0xde00c8, 0xc000115ec0, 0x0, 0xdd1740, 0xc0008e8990, 0x0, 0x0, 0x1)
```

cc @bcronin @adil-sultan


gdvalle (approved):
Thanks, Clay!

PR #69, opened 2022-02-03
metric dashboards: Remove y-axis

kayousterhout:
The y-axis was optional and is being removed from Lightstep's
public API, so remove support for it here.

### DIFF
--- a/lightstep/resource_metric_dashboard.go
+++ b/lightstep/resource_metric_dashboard.go
@@ -64,23 +64,6 @@ func getChartSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Computed: true,
 		},
-		"y_axis": {
-			Type:     schema.TypeList,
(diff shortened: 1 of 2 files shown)


lightstep/resource_metric_dashboard.go
```diff
@@ -64,6 +64,8 @@ func getChartSchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Computed: true,
 		},
+		// The y_axis is included here for backwards compatibility, but it's not used anywhere
+		// because it has been removed from Lightstep's public API.
```
smithclay, line 68: I'm thinking we can just remove from the schema and cut a new release version that's breaking (bump from 1.5 to 1.6) -- existing clients that haven't upgraded will still work (and will confirm in test).

kayousterhout:
Closing this PR because we decided not to change the public API

PR #73, opened 2022-02-18
Support API key as an attribute, not an environment variable name

dmathieu:
This adds an optional `api_key`, which allows specifying the API key value directly instead of having to rely on environment variables.
It is pretty common for terraform providers to be loaded with their attributes coming from a vault, and not have the values set as environment variables.

### DIFF
--- a/lightstep/provider.go
+++ b/lightstep/provider.go
@@ -3,17 +3,18 @@ package lightstep
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


dmathieu:
I believe the tests are failing because your CI doesn't like forks?

dmathieu:
I suppose the only way to fix it will be for you to cherry-pick the commit and reopen a new PR on a local branch.

smithclay:
Thanks again for the PR. This has been released in `v1.60.2`.

PR #83, opened 2022-05-26
Update the provider to use new input windows from lightstep public api

lambcode:
What does this PR do?
- Add 'timeseries_operator_input_window_ms' to metric queries and (both dashboards and metric alerts)
- Add 'final_window_operation' to single metric queries and formula metric queries (both dashboards and metric alerts)
- Remove 'evaluation_criteria', 'evaluation_window', 'num_sec_per_point' from metric alert expression as these are will be removed from the lightstep public API shortly.

Testing:
- Acceptance tests for a simple metric alerts was modified to use input windows for both dashboards and metric alerts

R @lightstep/team-alerts-be 
CC @smithclay


smithclay:
One minor thing: could you also update the `docs` with changes to the provider? Gets automatically updated in the Terraform registry after a release.

Also, just to confirm: since we're removing attributes, this is a breaking change and we'll release as a new minor version (say, `v1.61.0`) vs a patch.

client/metric_conditions.go
```diff
@@ -44,20 +41,28 @@ type Thresholds struct {
...
 
 type MetricQuery struct {
-	Metric             string        `json:"metric"`
-	Filters            []LabelFilter `json:"filters,omitempty"`
-	TimeseriesOperator string        `json:"timeseries-operator"`
-	GroupBy            GroupBy       `json:"group-by,omitempty"`
+	Metric                          string                `json:"metric"`
+	Filters                         []LabelFilter         `json:"filters,omitempty"`
+	TimeseriesOperator              string                `json:"timeseries-operator"`
+	TimeseriesOperatorInputWindowMs *int                  `json:"timeseries-operator-input-window-ms,omitempty"`
```
ltyson, line 58: just curious: is the `omitempty` part necessary, or just trying to document the fact that this field is optional?

docs/index.md
```diff
@@ -6,27 +6,21 @@ description: |-
   
 ---
 
-# Lightstep Provider
+# lightstep Provider
```
ltyson, line 9: Were these changes intentional?

docs/resources/metric_condition.md
```diff
@@ -33,6 +33,7 @@ resource "lightstep_metric_condition" "beemo-requests" {
     metric              = "requests"
     query_name          = "a"
     timeseries_operator = "delta"
+    timeseries_operator_input_window_ms = 30000
```
ltyson, line 36: nit: align the values vertically (like the go formatter would do in code)

client/metric_conditions.go
lambcode: I don't think it is strictly necessary. If the public api get a request that looks like
```
...
"timeseries-operator": "min",
"timeseries-operator-input-window-ms", null,
...
```
I'm pretty sure that would be functionaly equivelant to
```
...
"timeseries-operator": "min",
...
```
I used the omitempty to match the other optional fields in these structs like `Filters`

docs/resources/metric_condition.md
```diff
@@ -55,44 +56,41 @@ resource "lightstep_metric_condition" "beemo-requests" {
...
-- **operand** (String)
-- **thresholds** (Block List, Min: 1, Max: 1) (see [below for nested schema](#nestedblock--expression--thresholds))
+- `operand` (String)
+- `thresholds` (Block List, Min: 1, Max: 1) (see [below for nested schema](#nestedblock--expression--thresholds))
 
 Optional:
 
-- **is_multi** (Boolean)
-- **is_no_data** (Boolean)
-- **num_sec_per_point** (Number)
```
ltyson, line 87: what did this field do?

docs/index.md
lambcode: Weird. Not intentional, but this was autogenerated by the `tfplugindocs` command.

docs/resources/metric_dashboard.md
```diff
@@ -79,61 +80,88 @@ resource "lightstep_metric_dashboard" "customer_charges" {
...
-- **hidden** (Boolean)
-- **query_name** (String)
+- `hidden` (Boolean)
+- `query_name` (String)
+
+Optional:
+
+- `display` (String)
+- `exclude_filters` (List of Map of String) Not-equals filters (operand: neq)
+- `filters` (List of Map of String) Non-equality filters (operand: contains, regexp)
```
ltyson, line 125: Is this field new? It seems unrelated to the rest of your changes, and potentially redundant with `include_filters`?

docs/resources/metric_condition.md
lambcode: Oops sorry. Meant to comment on this. I have no idea what this is, but the public API does not accept or return this field. Made sense to me to remove it while we were making breaking changes.

docs/resources/metric_dashboard.md
lambcode: I did not change anything with filters, include_filters, etc. I just don't think the docs have been updated in a while, so the `tfplugindocs` command added it here. 

BTW, AFAICT `filters` is used for regex matching only.

lightstep/resource_metric_condition.go
```diff
@@ -212,8 +198,15 @@ func getQuerySchema(includeSpansQuery bool) map[string]*schema.Schema {
 			Type:         schema.TypeString,
 			Optional:     true,
 			Computed:     true,
-			ValidateFunc: validation.StringInSlice([]string{"rate", "delta", "avg", "last"}, false),
+			ValidateFunc: validation.StringInSlice([]string{"rate", "delta", "last", "min", "max", "avg"}, false),
+		},
+		"timeseries_operator_input_window_ms": {
+			Type:     schema.TypeInt,
+			Optional: true,
+			// Duration micros must be at least 30s and an even number of seconds
```
ltyson, line 206: "an event number of seconds" or "a multiple of 30s"

lightstep/resource_metric_condition.go
```diff
@@ -268,6 +261,30 @@ func getQuerySchema(includeSpansQuery bool) map[string]*schema.Schema {
...
+		Type:     schema.TypeList,
+		Optional: true,
+		MaxItems: 1,
+		MinItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"operator": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					ValidateFunc: validation.StringInSlice([]string{"min", "max", "avg"}, false),
```
ltyson, line 275: TODO: validate with Jake before merging?

lightstep/resource_metric_condition.go
```diff
@@ -574,6 +583,18 @@ func buildQueries(queriesIn []interface{}, includeSpansQuery bool) ([]client.Met
 			},
 		}
 
+		timeseriesOperatorInputWindowMs := query["timeseries_operator_input_window_ms"]
+		if timeseriesOperatorInputWindowMs != 0 { // 0 here indicates that the value was not set in terraform file
+			value := timeseriesOperatorInputWindowMs.(int)
+			newQuery.Query.TimeseriesOperatorInputWindowMs = &value
+		}
+
+		if newQuery.Type == "single" {
```
ltyson, line 592: Do we also need to check whether it's a metrics or spans alert to know whether to set `newQuery.Query` vs `newQuery.SpansQuery`? I think you mentioned that you were just opportunistically adding some support for spand queries, so feel free to just add a TODO if this comment isn't part of the current scope of work (which I think it's not)

lightstep/resource_metric_condition.go
lambcode: Approval has been given to go forward with 'avg' support in the final time window https://lightstep.slack.com/archives/C0242JADM5F/p1653597784428679

lightstep/resource_metric_condition.go
lambcode: I think you could technically specify 44s and qengine would not complain. 🤷

docs/index.md
ltyson: Sorry to be all "last one who touched it owns it" on you, but would you mind checking to see where this changed and try to verify whether it was intentional?

lightstep/resource_metric_condition.go
ltyson: Oh, so it literally does mean an even number of seconds? Why not an odd number??? This question is not relevant to this PR, I'm now just curious :)

lightstep/resource_metric_condition.go
ltyson: Oh yay! I'm just behind on slack 👍

lightstep/resource_metric_condition.go
lambcode: You could have an odd number of seconds. What I'm trying to get at here is that you can't specify fractions of seconds even though the unit is milliseconds. I can see now this needs rephrasing :)
45000 is valid
44000 is valid
1000 is invalid
44500 is invalid

lambcode: this is because windows in the qengine api are sepcified in seconds (type int)

lightstep/resource_metric_condition.go
lambcode: Correct. This PR is not attempting to support span alerts.

docs/index.md
lambcode: Looks like you can add doc template files to override some of this. These template files also allow us to keep the examples in the autogenerated file without needed to manually modify it after the fact. I'm going to add templates for the small set of files that we care about for this PR, and then follow up with another PR to fix the other doc files

lightstep/resource_metric_condition.go
mattcarmody: maybe "round" in place of "even" would squash the confusion? i.e. % 1000 == 0. or micros must represent a round integer in seconds

mattcarmody (approved):
lgtm

PR #88, opened 2022-06-13
LS-30243: Exclude heatmap display type from updating default latency_…

Dolphinsgrin:
What does this PR do?
-----------------------------
Creates an exception for `heatmap` display requests from terraform's desire to insert the default `latency_percentiles`.

Why are you doing it?
-----------------------------
Terraform currently "corrects" empty `latency_percentiles` by inserting all four [50, 95, 99, 99.9] by default. This is problematic for `heatmap` display requests, which does the inverse in the public-api: it deletes any/all `latency_percentiles` it encounters. We suspect that this correction/decorrection when terraform is applied is the cause of repeated `diff` responses during the `apply` phase.

R/CC: https://github.com/orgs/lightstep/teams/team-dashboards


lightstep/resource_metric_condition.go
```diff
@@ -490,11 +490,11 @@ func buildSpansGroupByKeys(keysIn []interface{}) []string {
 	return keys
 }
 
-func buildLatencyPercentiles(lats []interface{}) []float64 {
+func buildLatencyPercentiles(lats []interface{}, display string) []float64 {
 	var latencies []float64
 
-	// default
-	if len(lats) == 0 {
+	// default (heatmap default is not desired)
```
MisterSquishy, line 496: nit: maybe a higher level comment like "heatmap queries don't compute percentiles"

lightstep/resource_metric_condition.go
```diff
@@ -554,12 +554,13 @@ func buildQueries(queriesIn []interface{}, includeSpansQuery bool) ([]client.Met
...
 				}
+				display := query["display"].(string)
 				newQuery := client.MetricQueryWithAttributes{
 					Name:       query["query_name"].(string),
 					Type:       "spans_single",
 					Hidden:     query["hidden"].(bool),
-					Display:    query["display"].(string),
-					SpansQuery: buildSpansQuery(spansQuery),
+					Display:    display,
+					SpansQuery: buildSpansQuery(spansQuery, display),
```
MisterSquishy, line 563: do we need to do this for metrics queries too?

MisterSquishy (commented):
lgtm but i think youve gotta do the versioning dance again

lightstep/resource_metric_condition.go
Dolphinsgrin: The only place these latency values are inserted are in this file.

Dolphinsgrin: Specifically, a `Find in Files` for the value `99.9` returns only one hit.

Dolphinsgrin:
> lgtm but i think youve gotta do the versioning dance again

In progress. Good catch. This has to build and publish first (post-merge) before the lockfile and version update can occur in Lightstep proper.

lightstep/resource_metric_condition_test.go
```diff
@@ -363,8 +364,8 @@ func testAccCheckMetricConditionExists(resourceName string, condition *client.Me
 			return fmt.Errorf("ID is not set")
 		}
 
-		client := testAccProvider.Meta().(*client.Client)
-		cond, err := client.GetMetricCondition(context.Background(), test_project, tfCondition.Primary.ID)
+		providerClient := testAccProvider.Meta().(*client.Client)
+		cond, err := providerClient.GetMetricCondition(context.Background(), test_project, tfCondition.Primary.ID)
```
Dolphinsgrin, line 368: This is to make linter happy. `client` was colliding with an import. Renaming makes linter happy.

lightstep/resource_metric_condition.go
```diff
@@ -490,11 +490,11 @@ func buildSpansGroupByKeys(keysIn []interface{}) []string {
 	return keys
 }
 
-func buildLatencyPercentiles(lats []interface{}) []float64 {
-	var latencies []float64
+func buildLatencyPercentiles(lats []interface{}, display string) []float64 {
+	latencies := make([]float64, 0)
```
MisterSquishy, line 494: suuuper nit but it would be more efficient to `make([]float64, 0, len(lats))`

lightstep/resource_metric_condition.go
MisterSquishy: 👍 probably worth testing a metrics heatmap, but i see no reason it shouldnt work

MisterSquishy (approved):

lightstep/resource_metric_condition_test.go
```diff
@@ -784,3 +785,71 @@ func TestValidateGroupBy(t *testing.T) {
...
+			args: args{
+				lats:    []interface{}{},
+				display: "heatmap",
+			},
+			want: []float64{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := buildLatencyPercentiles(tt.args.lats, tt.args.display); !reflect.DeepEqual(got, tt.want) {
```
lambcode, line 850: `require.Equal` calls DeepEqual under the hood. I think it's more idiomatic to use `require.Equal` though in our unit tests.

Dolphinsgrin: Interesting. This boiler-plate is what IntelliJ creates when using `Generate test for function`. You're thinking like this?
`for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
	got := buildLatencyPercentiles(tt.args.lats, tt.args.display)
	require.Equal(t, got, tt.want, "buildLatencyPercentiles() = %v, want %v", got, tt.want)
    })
}`

lambcode: `require.Equal` will output the diff for you, so I think this could be simplified further to:
```
for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got := buildLatencyPercentiles(tt.args.lats, tt.args.display)
                         require.Equal(tt, got, tt.want)
                 }
}
```

PR #90, opened 2022-08-18
LS-29750 Update terraform provider to support span alerts

lambcode:
This change has two parts:
1) Update span queries in the provider to support input windows
2) Update the provider to allow span queries in metric alerts (aka span alerts)

Additionally:
- The documentation was updated
- Tests added to ensure that span alerts can be created

FYI, this repo is public.

R: @mattcarmody @ltyson @gontzess


lightstep/resource_metric_dashboard_test.go
```diff
@@ -220,14 +226,19 @@ resource "lightstep_metric_dashboard" "test" {
...
 				Config: spansQueryDashboardConfig,
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckMetricDashboardExists(resourceNameSpans, &dashboard),
-					resource.TestCheckResourceAttr(resourceNameSpans, "dashboard_name", "Acceptance Test Dashboard"),
-					resource.TestCheckResourceAttr(resourceNameSpans, "chart.0.query.0.display", "line"),
-					resource.TestCheckResourceAttr(resourceNameSpans, "chart.0.query.0.hidden", "false"),
-					resource.TestCheckResourceAttr(resourceNameSpans, "chart.0.query.0.spans.0.operator", "error_ratio"),
-					resource.TestCheckResourceAttr(resourceNameSpans, "chart.0.query.0.spans.0.query", "service IN (\"frontend\")"),
-				),
+				Check: func(state *terraform.State) error {
```
mattcarmody, line 229: why use a closure here? I don't see state being used

mattcarmody (approved):
sorry for the delay, fell through the cracks! LGTM (where M is not well versed in terraform)

lightstep/resource_metric_dashboard_test.go
lambcode: I added this so I could set a breakpoint while debugging. Will remove!

PR #91, opened 2022-08-19
LS-33927 Support span formulas

lambcode:
This PR supports span formulas in the terraform provider. If the set of queries for a dashboard/alert contain any span single queries the formula gets the type `spans_composite` which is needed for span formulas.

R: @lightstep/team-alerts-be


lightstep/resource_metric_condition.go
```diff
@@ -650,6 +652,15 @@ func buildQueries(queriesIn []interface{}) ([]client.MetricQueryWithAttributes,
 		}
 		newQueries = append(newQueries, newQuery)
 	}
+
+	if hasSpanSingle {
+		for i := 0; i < len(newQueries); i++ {
+			if newQueries[i].Type == "composite" {
+				newQueries[i].Type = "spans_composite"
+			}
```
mattcarmody, line 660: will things get weird if there's a mix of metrics and spans queries?

lightstep/resource_metric_condition.go
```diff
@@ -561,6 +562,7 @@ func buildQueries(queriesIn []interface{}) ([]client.MetricQueryWithAttributes,
 
 		spansQuery := query["spans"]
 		if spansQuery != nil && len(spansQuery.([]interface{})) > 0 {
+			hasSpanSingle = true
```
mattcarmody, line 565: not new code, but on 564 is there any concern of the type assertion panicking?

mattcarmody (commented):
a couple questions

lightstep/resource_metric_condition.go
lambcode: AFAIK, you can't actually mix span and metric queries. I just double checked this is the case in the UI. I think if we wanted to correctly fix this in the future we would need to not have a different composite query type for spans and metrics. They have no user facing difference anyway.

lightstep/resource_metric_condition.go
lambcode: I think this is fine because of the provider type that we defined ensures this will be an array. Honestly though, I'm guessing here. Seeing as it hasn't bitten us yet, I'm not too worried about this.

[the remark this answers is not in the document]
lambcode: Good idea!

lightstep/resource_metric_condition.go
mattcarmody: I agree someone shouldn't be able to and the UI prevents it but can someone write a terraform file to mix them? Then my concern is here we override one of the types in a way that might result in us sending in weird data that makes it through the public API when it shouldn't

lambcode: I just tried it and it fails validation on the server side if you try and mix metric and span queries in a formula.

mattcarmody: thank you for checking that!

mattcarmody (approved):
LGTM

PR #92, opened 2022-08-22
removes the Computed flag from some query flags that are not computed

MisterSquishy:
i noticed that if you delete the `metric` field from a metric alert, terraform sees no diff. all of these usages of `Computed` seem wrong to me, as per https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#computed... am i missing something?

### DIFF
--- a/lightstep/resource_metric_condition.go
+++ b/lightstep/resource_metric_condition.go
@@ -179,7 +179,6 @@ func getQuerySchema(includeSpansQuery bool) map[string]*schema.Schema {
 		"metric": {
 			Type:     schema.TypeString,
 			Optional: true, // optional for composite formula
-			Computed: true,
 		},
 		"hidden": {
 			Type:     schema.TypeBool,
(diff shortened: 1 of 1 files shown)


MisterSquishy:
test failure:
```
    resource_metric_dashboard_test.go:180: Step 3/5 error: Error running apply: exit status 1
        
        Error: Failed to update metric dashboard: status 400 (400 Bad Request): "{\"errors\":[\"rpc error: code = InvalidArgument desc = Invalid chart type () - try 'timeseries'\"]}"
        
          with lightstep_metric_dashboard.test,
          on terraform_plugin_test.tf line 2, in resource "lightstep_metric_dashboard" "test":
           2: resource "lightstep_metric_dashboard" "test" {
```

how is that even related??? that type field is defined on a [totally different resource](https://github.com/lightstep/terraform-provider-lightstep/blob/df6dec151e79a08a9ef71f051c706d5f363cc6cb/lightstep/resource_metric_dashboard.go#L60-L64), and [is specified](https://github.com/lightstep/terraform-provider-lightstep/blob/cde0a031ff651379e66fcac8c0b711a59f827531/lightstep/resource_metric_dashboard_test.go#L63) in the failing test. i definitely think im missing something here.

PR #98, opened 2022-09-28
bcronin/query string

bcronin:
# What is this change?

Does the brunt of the update of the Lightstep Terraform Provider to support UQL.

> **Note: this PR is intended to be merged into the `v1.70-dev` branch until v1.70 is ready to be published.**

-   Bumps the version from 1.62.0 to 1.70.0 as this adds functionality _without_ deprecating prior query formats (we can do that in a v2.0 release)
-   Adds a new Terraform resource `lightstep_dashboard` which is similar to `lightstep_metric_dashboard` but _only_ accepts `query_string` as the query definition. This is intended to help future migration to a v2.0 where `lightstep_metric_dashboard` can be deprecated and replaced by the unified `lightstep_dashboard` resource
-   Deprecates the `tql` field on `lightstep_metric_dashboard`. The field is replaced with a `query_string` field
-   Updates the docs and tests with basic examples

Internal changes:

-   Renames some internal classes and functions from `Metric` to `Unified` in the cases where the code does not apply only to metrics
-   Addresses some distracting lint warnings about Go error capitalization


DEVELOPMENT.md
```diff
@@ -5,35 +5,35 @@
 If you're contributing changes or code to the provider, the integration tests create, update, and destroy real resources in a Lightstep-managed integration environment.
 
 To run the tests, first get an API key with a Member role for the public environment and run:
+
```
bcronin, line 8: Note the changes to DEVELOPMENT.md are cosmetic, Markdown-linting changes only.

Makefile
```diff
@@ -33,8 +33,16 @@ lint:
 
 .PHONY: acc-test
 acc-test:
+ifndef LIGHTSTEP_API_KEY_PUBLIC
+	$(error LIGHTSTEP_API_KEY_PUBLIC must be defined for acc-test)
+endif
```
bcronin, line 38: Added this as it is a little confusing that `LIGHTSTEP_API_KEY_PUBLIC` needs to be set, but the internal error says `LIGHTSTEP_API_KEY` needs to be set (since, per the next line, the environment variable is remapped before execution).

README.md
```diff
@@ -36,19 +37,21 @@ module "aws-dashboards" {
...
 ```
    # Avoids 500 errors when creating many resources.
-   terraform apply -parallelism=1 
+   terraform apply -parallelism=1
 ```
+
 ## Development
 
-See `DEVELOPMENT.md`.
+See [`DEVELOPMENT.md`](DEVELOPMENT.md).
```
bcronin, line 54: Added the link here. All other changes to this file are cosmetic, Markdown-linting changes.

client/metric_dashboards.go
```diff
@@ -7,18 +7,18 @@ import (
 	"net/http"
 )
 
-type MetricDashboard struct {
-	Type       string                    `json:"type"`
-	ID         string                    `json:"id"`
-	Attributes MetricDashboardAttributes `json:"attributes,omitempty"`
+type UnifiedDashboard struct {
```
bcronin, line 10: This file has only renaming changes to avoid the confusing scenario where we would otherwise be calling a function or using a struct with "metric" in the name even though it applies to any dashboard.

lightstep/resource_dashboard.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func getQueriesFromDashboardResourceData(queriesIn []client.MetricQueryWithAttributes) []interface{} {
+	var queries []interface{}
+	for _, q := range queriesIn {
+		qs := map[string]interface{}{
+			"hidden":       q.Hidden,
+			"display":      q.Display,
+			"query_name":   q.Name,
+			"query_string": q.TQLQuery,
```
bcronin, line 140: This file is _mostly_ a clone of `resource_metric_dashboard.go` with the primary change here being the resource data contains _only_ `query_string`.  I thought forking the code would be a cleaner, easier path than refactoring the metric dashboard code with conditionals.

lightstep/resource_dashboard.go
```diff
@@ -0,0 +1,218 @@
...
+	d.SetId(id)
+	if err := setResourceDataFromDashboard(project, *dash, d); err != nil {
+		return nil, fmt.Errorf("failed to set dashboard from API response to terraform state: %v", err)
+	}
+
+	return []*schema.ResourceData{d}, nil
+
+}
+
+func getUnifiedQuerySchema() map[string]*schema.Schema {
```
bcronin, line 197: This function is also unique to the `lightstep_dashboard` resource.

lightstep/resource_metric_condition.go
```diff
@@ -257,11 +257,19 @@ func getQuerySchema() map[string]*schema.Schema {
...
+			Optional:    true,
+			Computed:    true,
 		},
 		"spans": getSpansQuerySchema(),
+		"query_string": {
+			Type:        schema.TypeString,
+			Description: "query expressed in Lightstep Unified Query Language (UQL)",
+			Optional:    true,
+			Computed:    true,
+		},
```
bcronin, line 272: I was on the fence about whether to add `query_string` to this resource or simply have the deprecation message say "use the lightstep_dashboard resource instead". I decided to provide both, but am glad to change that if there's a strong preference.

lightstep/resource_metric_dashboard.go
```diff
@@ -120,50 +135,50 @@ func resourceMetricDashboardRead(ctx context.Context, d *schema.ResourceData, m
...
 	if err := setResourceDataFromMetricDashboard(d.Get("project_name").(string), *dashboard, d); err != nil {
-		return diag.FromErr(fmt.Errorf("Failed to set metric dashboard from API response to terraform state: %v", err))
+		return diag.FromErr(fmt.Errorf("failed to set metric dashboard from API response to terraform state: %v", err))
 	}
 
 	return diags
 }
 
-func getMetricDashboardAttributesFromResource(d *schema.ResourceData) (*client.MetricDashboardAttributes, error) {
+func getUnifiedDashboardAttributesFromResource(d *schema.ResourceData) (*client.UnifiedDashboardAttributes, error) {
```
bcronin, line 155: I didn't move this out of `resource_metric_dashboard.go` so it would be clear it has only been renamed, not functionally changed. I can move it in a follow-up.

lightstep/resource_metric_condition.go
MisterSquishy: the addition of this field and the deprecation of the `tql` field are the only changes needed to address the "it's-not-called-tql-anymore" problem, right? 

the creation of the `lightstep_dashboard` resource addresses a different problem (the use of `metrics` when we mean `unified`). i think the latter problem is also worth fixing, but it might've been good to split into a separate PR; this one is a lot to review

MisterSquishy: what's the plan for releasing v2? i have some concern about the `lightstep_dashboard` resource, just because it leaves us in a hard-to-maintain duplicated state until we get around to removing `lightstep_metric_dashboard`. but if you're planning on doing that real soon then it's probably not much of a concern

it's nice that folks get to migrate to v2-compatible hcl before we actually release it, but in practice i dont think `s/lightstep_metric_dashboard/lightstep_dashboard` and `s/tql/query_string` represent huge migration burdens for folks looking to upgrade when the time comes. did you consider just adding `lightstep_dashboard` and removing `lightstep_metric_dashboard` as part of the v2 breaking changes?

lightstep/resource_dashboard.go
MisterSquishy: outside of [this one](https://github.com/lightstep/terraform-provider-lightstep/pull/98/files#diff-c9ec685f96ccaad128fe473a95c4e4bbea0ba87055d41929e5743cbd59866ac6R558-R561), which other conditionals would we need?

MisterSquishy (commented):
i think this all looks good! i left a few nonblocking questions, there's just one documentation thing i think needs updating

docs/resources/dashboard.md
```diff
@@ -0,0 +1,104 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "lightstep_metric_dashboard Resource - terraform-provider-lightstep"
+subcategory: ""
+description: |-
+
+---
+
+# lightstep_metric_dashboard (Resource)
```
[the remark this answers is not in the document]
bcronin: Good catch. Fixed.

lightstep/resource_dashboard.go
bcronin: Maybe "refactoring with conditionals" was not the right way to express what I thinking. Given the plan is to completely deprecate the lightstep_metric_dashboards resource relatively, it just seemed simpler to fork the code so it's clear what needs to be kept up to data and what needs to be deleted. If the two resources are sharing the same code, coming from the outside, it'd be a little confusing to me what was being used by which resource otherwise.  I don't feel strongly about this as I was mostly just assuming the old code will be deprecated relatively soon.

lightstep/resource_metric_condition.go
bcronin: A bunch of thoughts...

(1) I don't think this creates a  "hard-to-maintain duplicated state" as we shouldn't be updating `lightstep_metric_dashboard` anymore. I suspect the forked file duplicated code is giving the impression this is more complex than it is (the "resources" are shallow interfaces on the same underlying `Client` code).  I'll consider removing the forking to see if I can make it more clear this is basically a renaming and TF fields/state clean-up.

(2) From backwards compatibility work in the past and do have a somewhat strong opinion that the "better" way to do binary breaks is to introduce the new interface first, have a period of overlap with both available (basically the deprecation period), then removing the old one. I don't think removing an old interface and introducing a new one at the same time makes for as nice a developer experience as it's harder to do incremental migrations or test configurations with partial migrations (e.g. move one dashboard, run it side-by-side, make sure it didn't change anything).  Also, it's a bit more than "just a string substitution" since the resource changes the user's underlying Terraform state.

(Aside: I didn't mark _all_ the old lightstep_metric_dashboard query fields as deprecated since I'm not sure exactly how Terraform treats deprecated fields, but it _seems_ like it might force the user to fix all those before `apply` can be run again -- and per my point about being able to migrate w/o having to do it wholesale, I wanted to avoid that.)

MisterSquishy: 👍 this all sounds good!

MisterSquishy (approved):

PR #99, opened 2022-09-29
Revert query_string addition to lightstep_metric_dashboard

bcronin:
Why are you making this change?
---------

Reverts the updates to include an optional “query_string” field in the lightstep_metric_dashboard schema. The schema change was causing a number of “false positive” updates in `lightstep_metric_dashboard` resources. Reverting the change until the cause of that is better understood.

Also fixes a conversion cast that should have been checking for nil.


.go-version
```diff
@@ -1 +1 @@
-1.70.0
\ No newline at end of file
+1.70.2
```
MisterSquishy, line 1: `1.70.1`?

lightstep/resource_metric_condition.go
```diff
@@ -555,9 +549,9 @@ func buildQueries(queriesIn []interface{}) ([]client.MetricQueryWithAttributes,
 	for _, query := range queries {
 
 		// When checking if this chart uses a query string, check deprecated TQL field as well
-		queryString := query["query_string"].(string)
-		if queryString == "" {
-			queryString = query["tql"].(string)
+		queryString, ok := query["query_string"].(string)
```
bcronin, line 552: This is still needed as the unified dashboard code calls this as well

MisterSquishy (approved):

.go-version
bcronin: I figured patch-level version numbers are free and this doesn't really make a difference :)

PR #103, opened 2022-10-04
Add export support for `lightstep_dashboard` resource

bcronin:
### Why are you making this change?

This PR adds export support for the newly added `lightstep_dashboard` resource.

This additionally fixes some string export issues around special characters:
* Query strings containing newline characters are now exported using the [heredoc notation](https://www.terraform.io/language/expressions/strings#heredoc-strings
* Likewise query strings containing quotes, tabs, or other characters requiring escaping will also use the heredoc notation to simplify the user editing/diffing experience
* Single-line query strings that do not require escaping retain the single-line representation

Worth noting: the provider does not currently support dashboards using a mix of the legacy query format and query-string format charts. This would require a new resource type. Users are encouraged to migrate charts to the query-string-based `lightstep_dashboard` resource instead.


exporter/exporter.go
```diff
@@ -57,8 +59,106 @@ resource "lightstep_metric_dashboard" "exported_dashboard" {
...
+      hidden              = {{.Hidden}}
+      query_string        = {{escapeQueryString .TQLQuery}}
+	}
+{{end}}
+  }
+{{end}}
+}
+`
+
+func escapeHCLString(input string) string {
```
bcronin, line 84: Replaces `escapeSpanQuery` which was only escaping double-quote.  There weren't any bug logged on this AFAIK, but the change seems correct.

exporter/exporter.go
```diff
@@ -57,8 +59,106 @@ resource "lightstep_metric_dashboard" "exported_dashboard" {
...
+			}
+		}
+	}
+	if hasLegacyQuery {
+		return false, nil
+	}
+	return true, nil
+}
+
+func exportToHCL(wr io.Writer, d *client.UnifiedDashboard) error {
```
bcronin, line 135: Split this worker function out of `Run` so the export could be unit tested more easily. This also adds logic to check which resource type should be used based on the kinds of queries in the charts.

exporter/exporter.go
```diff
@@ -57,8 +59,106 @@ resource "lightstep_metric_dashboard" "exported_dashboard" {
...
+  chart {
+    name = "{{.Title}}"
+    rank = "{{.Rank}}"
+    type = "{{.ChartType}}"
+{{range .MetricQueries}}
+    query {
+      query_name          = "{{.Name}}"
+      display             = "{{.Display}}"
+      hidden              = {{.Hidden}}
+      query_string        = {{escapeQueryString .TQLQuery}}
```
MisterSquishy, line 76: did you check that this works without the heredoc syntax? e.g. if my UQL query is just `metric cpu.utilization | delta` (no newlines), i think this will output `query_string = metric cpu.utilization | delta` -- i have no idea if terraform will be upset about the lack of quotes

exporter/exporter.go
```diff
@@ -26,13 +28,13 @@ resource "lightstep_metric_dashboard" "exported_dashboard" {
...
       spans {
-         query         = "{{escapeSpanQuery .SpansQuery.Query}}"
+         query         = "{{escapeHCLString .SpansQuery.Query}}"
          operator      = "{{.SpansQuery.Operator}}"
          group_by_keys = [{{range .SpansQuery.GroupByKeys}}"{{.}}",{{end}}]{{if eq .SpansQuery.Operator "latency"}}
          latency_percentiles = [{{range .SpansQuery.LatencyPercentiles}}{{.}},{{end}}]{{end}}
       }
 {{end}}{{if .TQLQuery}}
-      tql                 = "{{.TQLQuery}}"
+      tql                 = "{{escapeQueryString .TQLQuery}}"
```
MisterSquishy, line 37: wont these quotes get in the way if we're using the heredoc syntax?

exporter/exporter.go
```diff
@@ -57,8 +59,106 @@ resource "lightstep_metric_dashboard" "exported_dashboard" {
...
+		return escapeHCLString(input)
+	}
+}
+
+// dashboardUsesQueryString returns true if any chart in the dashboard
+// uses the query string format.
+//
+// The Provider does not support mixing legacy query charts and
+// query-string charts in the same dashboard and will return an error
+// if there is a mix.
```
MisterSquishy, line 112: is there a reason we couldnt just return the legacy format (`lightstep_metric_dashboard`) in this case? it will use the `tql` field name for the query strings, but that seems better than just failing

exporter/exporter.go
MisterSquishy: on the other hand, if we decide to never output `lightstep_metric_dashboard`s when using TQL queries with this tool, then i think we'll never actually need this change?

exporter/exporter_test.go
```diff
@@ -0,0 +1,88 @@
...
+		QueryString string
+		Expected    string
+	}{
+		{
+			QueryString: "",
+			Expected:    "",
+		},
+		{
+			QueryString: "metric requests | rate 10m",
+			Expected:    "query_string        = metric requests | rate 10m",
```
MisterSquishy, line 29: this is the kind of thing i mentioned about -- do we know terraform won't complain about the lack of quotes here?

exporter/exporter.go
bcronin: Good catch! That's definitely wrong and needs to be fixed!

exporter/exporter.go
```diff
@@ -57,8 +59,106 @@ resource "lightstep_metric_dashboard" "exported_dashboard" {
...
+	// Use "heredoc" syntax if the query contains any newlines or other characters that'd
+	// need to be escaped and make the single line representation less convenient to work with.
+	if strings.Contains(input, "\"") ||
+		strings.Contains(input, "\n") ||
+		strings.Contains(input, "\r") ||
+		strings.Contains(input, "\t") ||
+		strings.Contains(input, "\\") {
+		return "<<EOT\n" + input + "\nEOT"
+	} else {
+		return escapeHCLString(input)
```
MisterSquishy, line 103: given that we're checking for all the same things, don't we know that `escapeHCLString` won't actually change `input`? would it be clearer to just return `input`?

exporter/exporter.go
bcronin: Yes, it would work.  My assumption was (1) we're trying to get folks off the legacy format and (2) in our internal usage, there are no such mixed dashboards.  I thought pushing for "please migrate" would be better than perpetuating the deprecated `tql` field.  I don't feel strongly about this though.

MisterSquishy: i think we have some of this internally... e.g. the [machine dashboards](https://app-meta.lightstep.com/lightstep-public/dashboard/alertevaluator-machine-terraform/wZh3WZpp) mix [legacy](https://github.com/lightstep/lightstep/blob/main/terraform/lightstep/lightstep/dashboards_machine.tf#L27-L32) and [query string](https://github.com/lightstep/lightstep/blob/main/terraform/lightstep/lightstep/dashboards_machine.tf#L314-L321). i'm mostly just worried that, as it stands today, "please migrate" means they have to have all their queries stored in editor mode, which may not be the experience they want. as a user, id probably just say "dang, guess i can't use the exporter" rather than "ok, ill change all my queries"

exporter/exporter.go
bcronin: Good catch

exporter/exporter.go
bcronin: Ah, I didn't realize we had mixed dashboards already.  In that case, yes, this should use the legacy format...I'll update the PR.

MisterSquishy (approved):

PR #104, opened 2022-10-11
adds span count as a valid operator

MisterSquishy:
the lightstep UI has been updated to include `count` as a valid operator for spans queries. this updates the terraform provider to allow such queries as well.


MisterSquishy:
@lambcode do we expect these errors from the acceptance tests?

```
=== RUN   TestAccStreamAlertingRule
    resource_alerting_rule_test.go:102: Step 2/3 error: Error running apply: exit status 1
        
        Error: failed to create stream condition: status 410 (410 Gone): "Stream alerts are no longer supported, use metric alert endpoints\n"
        
          with lightstep_stream_condition.beemo_errors,
          on terraform_plugin_test.tf line 9, in resource "lightstep_stream_condition" "beemo_errors":
           9: resource "lightstep_stream_condition" "beemo_errors" {
```

lambcode:
Hmm. I guess that the project used for these tests has been converted to no longer support stream alerts. I think the easiest thing to do here is just delete those tests.

bcronin (approved):
I need to defer to @lambcode on the alert test failures.

For the code change itself, looks good to me!

lightstep/resource_metric_condition.go
```diff
@@ -136,7 +136,7 @@ func getSpansQuerySchema() *schema.Schema {
 				},
 				"operator": {
 					Type:         schema.TypeString,
-					ValidateFunc: validation.StringInSlice([]string{"latency", "rate", "error_ratio"}, false),
+					ValidateFunc: validation.StringInSlice([]string{"latency", "rate", "error_ratio", "count"}, false),
```
lambcode, line 139: Should we just remove this validation altogether and rely on the validation in the LS SaaS?

lambcode (approved):

lightstep/resource_metric_condition.go
MisterSquishy: yeah that's a fair question... i think i wanna punt, cuz there's a fair number of places that do this and i'd prefer to be comprehensive if/when we change this paradigm. but agreed that there's not much upside (and a lot of downside) to duplicating this kind of validation logic across the public api and the terraform provider

MisterSquishy: https://lightstep.atlassian.net/browse/LS-36604

bcronin: Not 100% confident in this, but I assume the ValidateFunc implementation would result in a more "ergonomic" error message to the user as Terraform itself will understand the semantics of the error better than opaque API response. That would be the argument for keeping the TF validation as well (but am not sure it is worth the added maintenance of a bit of duplicate code -- so I do not have a strong opinion here!).

PR #107, opened 2022-11-10
Provide error message when lightstep_dashboard cannot import chart

bcronin:
## What does this do?

Fixes a bug where `terraform import` of a dashboard containing a chart (or charts) containing queries that could not be represented would silently and incorrectly _succeed_ with an empty query string, rather than producing an error.  It now produces an error message and the import will fail.

image

Specifically, this happens when the new `lightstep_dashboard` resource is used on a dashboard containing "legacy" queries that are not in the query string format. 

Using the `lightstep_metric_dashboard` resource is a viable workaround. A proper fix should be added for this scenario before `lightstep_metric_dashboard` is removed from the provider because of cases such as this.  A fuller fix to automatically convert the query type (rather than produce an error) is currently blocked until there is public API support for such an operation.


Makefile
```diff
@@ -1,35 +1,41 @@
+# Currently always builds for amd64
+#PLATFORM_ARCH=$(shell uname -m | tr '[:upper:]' '[:lower:]')
+PLATFORM_ARCH=amd64
+PLATFORM_NAME=$(shell uname -s | tr '[:upper:]' '[:lower:]')
+VERSION_TAG=$(shell cat .go-version)
+
 .PHONY: build
 build:
-	@go build -o terraform-provider-lightstep_v$(shell cat .go-version)
+	GOARCH=$(PLATFORM_ARCH) go build -o terraform-provider-lightstep_v$(VERSION_TAG)
```
bcronin, line 9: Historically it looks like we've always hard-coded the local build to `amd64`. The folders and filenames have been hard-coded to that.  However, it was not setting a flag on the go compiler so binaries were being produced for "amd64" even though they were compiled as "arm64" on my M1.  This fixes that.

The rest was a bit of clean-up to use named variables instead of repeated shell commands.

Lastly, I removed some of the `@` prefixes to not silence the Make commands. Maybe that's subjective, but I find it frustrating when a make command fails to have no output of what it was running!

lightstep/resource_dashboard.go
```diff
@@ -30,9 +32,33 @@ func getUnifiedQuerySchema() map[string]*schema.Schema {
...
+) ([]interface{}, error) {
+
 	var queries []interface{}
 	for _, q := range queriesIn {
+		// Check if a query is using the legacy representation.  In that case,
+		// it can't be represented in the "query_string" field and the call should
+		// fail. In the future, we should attempt to automatically convert legacy -> query
+		// string format. At the moment, there's no public API to implement this, so
+		// at least provide a clarifying error message.
+		if q.Type != "tql" {
```
MisterSquishy, line 48: my memory is failing me, is this always what the public api calls query types? it doesnt try to rename this as `uql` or `query_string` or anything? i can also go check but figured you might know offhand

MisterSquishy (approved):
one lil q but lgtm!

lightstep/resource_dashboard.go
bcronin: Per our offline, conversation, it should be "tql" at least for this version of the API (i.e. we'd need to update the provider anyway if it changed).

PR #110, opened 2022-11-17
[LS-38653] add lightstep notebook provider support

jdcrouse:
## This PR
adds terraform provider support for Notebooks CRUD based on the schemas defined in the Notebooks Public API

The Notebooks Public API has yet to be merged, so this cannot be merged or tested. 
Dependent on https://github.com/lightstep/lightstep/pull/36342 and others

### DIFF
--- a/client/notebooks.go
+++ b/client/notebooks.go
@@ -1,0 +1,141 @@
+package client
+
+import (
+	"context"
+	"encoding/json"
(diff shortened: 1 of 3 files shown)


MitchellDumovic:
Let's wait on this for now, given that we're not sure that we want to go ahead with terraform for these APIs.

PR #111, opened 2022-11-19
Convert lightstep_resource dashboards to query_string format on read

bcronin:
## What is this PR?

The recently introduced `lightstep_dashboard` resource is intended to exclusively use the `query_string` representation for queries.  This PR updates the read of lightstep_dashboard resources to implicitly convert any legacy format queries into query_string format. 

This includes the import and export functionality.  To keep backward compatibility with the export functionality, the new "resource type" specified on the command-line should be "lightstep_dashboard" (rather than just "dashboard").


MisterSquishy (approved):
🥳

PR #115, opened 2023-01-03
Add missing description field when creating a metric condition

mattcarmody:
The `Description` field wasn't being sent with requests to create a metric condition. I updated the version and acceptance test to accompany it. This is my first terraform provider change so if there's anything else I need to do let me know!


client/metric_conditions.go
```diff
@@ -119,6 +119,7 @@ func (c *Client) CreateMetricCondition(
 			Expression:    condition.Attributes.Expression,
 			Queries:       condition.Attributes.Queries,
 			AlertingRules: condition.Attributes.AlertingRules,
+			Description:   condition.Attributes.Description,
```
MisterSquishy, line 122: 🔥 looks like the [acceptance test](https://github.com/lightstep/terraform-provider-lightstep/blob/17f8d52e5a23b9099dc3eaaaf5d3cab5d2189fb2/lightstep/resource_metric_condition_test.go) never exercised this bit; mind updating those test cases to test descriptions?

MisterSquishy (commented):
dont forget to increment the [version](https://github.com/lightstep/terraform-provider-lightstep/blob/main/.go-version)!

client/metric_conditions.go
mattcarmody: I added a description to a create and update! it looks like all the acceptance tests are failing though, maybe with a permissions issue.. any ideas as to why?

MisterSquishy: hmmm my gut says [this dude](https://github.com/lightstep/terraform-provider-lightstep/blob/main/.github/workflows/pr.yml#L109) expired, but i lack permissions to update it 😬 i think we have to find someone powerful enough to update this secret

.go-version
```diff
@@ -1 +1 @@
-1.70.9
+1.70.10
```
MisterSquishy, line 1: ruh roh looks like someone [stole this version](https://github.com/lightstep/terraform-provider-lightstep/releases/tag/v1.70.10) and forgot to update

MisterSquishy (approved):
🔥 once we get that API token rotated these tests oughta pass; lgtm minus the version thingy

MisterSquishy:
(also fyi that creating/pushing a tag is not automated for some reason, you'll have to tag your `main` commit manually post-merge to trigger a build)

.go-version
gdvalle: this was me. I wasn't sure exactly what the release steps should be. I only merged PR and pushed up a lightweight tag, so if this is all that's missing perhaps we can drop a blurb in the README/DEVELOPING.md for the next passerby. Is there anything else? What's this file for if we don't need it to publish?

mattcarmody:
@MisterSquishy can I call on you to take a look at the failing dashboard acceptance tests? I'm guessing TQL validation has been tightened in some way since they were created

.go-version
MisterSquishy: yeah tbh i have no idea how this is _supposed_ to work... i think this file is just for bookkeeping so we don't have to remember which tags have been pushed already? but it would also be nice if merging a change to this file pushed a tag automatically.

it does feel like a quick update to the README is the easiest thing to do here, but itll still be super easy to forget this step

MisterSquishy:
@mattcarmody this is almost certainly fixed in https://github.com/lightstep/lightstep/pull/37573, though i have no idea why it would just start failing now. i think it's safe to ignore and ill check on it after that fix gets deployed

client/metric_conditions.go
mattcarmody: done!

PR #116, opened 2023-01-18
Add dependency_map as a valid chart display type

wfro:
Adds `dependency_map` as a valid display type for dashboard charts.

Had one question around `metric_condition`: I also updated the display type here but wasn't 100% sure dependency maps are relevant there - from context they're related to alerts but not sure if they're specifically a query we'd alert on (and DM wouldn't make much sense there), following up on this one.

<details>
<summary>test config for posterity</summary>

```
terraform {
  required_providers {
    lightstep = {
      source = "lightstep/lightstep"
      version = "1.70.12"
    }
  }
}

provider "lightstep" {
  api_key      = "<api key>"
  organization = "LightStep"
}

resource "lightstep_dashboard" "foo" {
  project_name   = "demo"
  dashboard_name = "dev_test2"
  dashboard_description = "Foo"

  chart {
    name = "Requests by Project"
    rank = 1
    type = "timeseries"

    query {
      hidden         = false
      query_name     = "a"
      display        = "line"
      query_string   = "metric requests | rate 10m | group_by [project_id], sum"
    }
  }

  chart {
    name = "DM"
    rank = 2
    type = "timeseries"

    query {
      hidden         = false
      query_name     = "b"
      display        = "dependency_map"
      query_string   = "spans_sample service = apache | assemble"
    }
  }
}
```
</details>


lightstep/resource_metric_condition.go
```diff
@@ -194,7 +194,7 @@ func getMetricQuerySchema() map[string]*schema.Schema {
 		"display": {
 			Type:         schema.TypeString,
 			Optional:     true,
-			ValidateFunc: validation.StringInSlice([]string{"line", "area", "bar", "big_number", "heatmap"}, false),
+			ValidateFunc: validation.StringInSlice([]string{"line", "area", "bar", "big_number", "heatmap", "dependency_map"}, false),
```
MisterSquishy, line 197: so there's a very unsatisfying answer to your question: NO `dependency_map` is not a valid display type for alert queries but YES it needs to be added here -- im pretty sure some legacy dashboards TF validation code calls through here

i have no idea what would happen if someone tried to make an alert with display type `dependency_map`... in the UI we only allow `line`. maybe there's some other validation code in here somewhere that does that validation for us? but regardless this is not a new problem (for instance heatmap makes just as little sense in an alert)

wfro:
Re docs: I didn't update the provider docs since it looks like they have more general examples and link out to LS api docs for specifics (where `dependency_map` is included as a display type), open to suggestions though if they can/should be improved. I think I'll need to publish a new version regardless though? Since `dependency_map` would fail validation as a `display` value during plan/apply current

lightstep/resource_metric_condition.go
bcronin: I also wonder what happens with `big_number` given we deprecated that and it's `big_number_v2` now...

PR #117, opened 2023-01-28
Add options for dependency map dashboard charts

wfro:
* Adds an optional `dependency_map_options` attribute to the `dashboard` resource for users defining a chart query with display = "dependency_map". 
* Updates the exporter to handle existing dashboards with dependency map charts

Not necessarily specific to this change, but it's worth noting that we only support UQL queries with `display = "dependency_map`. For dashboards with dependency map charts using the query builder, the exporter already handles converting those to UQL (yay).


lightstep/resource_dashboard.go
```diff
@@ -28,6 +28,25 @@ func getUnifiedQuerySchema() map[string]*schema.Schema {
 			Type:     schema.TypeString,
 			Required: true,
 		},
+		"dependency_map_options": {
+			Type:     schema.TypeList,
```
wfro, line 32: Was originally hoping to use [`TypeMap`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typemap) here but because it's not possible to declare specific string keys/validate values (afaict) I went with a pattern I found elsewhere in the provider that uses a `TypeList` with one element.

.go-version
```diff
@@ -1 +1 @@
-1.70.12
+1.70.13
```
Dolphinsgrin, line 1: Is this intentional?

Dolphinsgrin (approved):
Everything looks good! Minor concern about slipping a go-version change into the effort.

.go-version
bcronin: This seems correct. We should be bumping the version on every change.

bcronin (approved):

.go-version
wfro: Yeah, iirc this isn't used anywhere for official releases but is used for local builds during development

bcronin: FYI, this is used for building the official images: https://github.com/lightstep/terraform-provider-lightstep/blob/main/Makefile#L5

PR #122, opened 2023-02-07
Check type assertions and added GetStatusCode

heidmotron:
Added type assertion checks.
Added `GetStatusCode` method to handle nil check for empty response.

The above was added to handle panics in the Provider from type assertion mismatch and when the `http.Response` is nil.


smithclay (approved):
LGTM, nice tests :shipit:

client/stream_dashboards.go
```diff
@@ -114,8 +114,8 @@ func (c *Client) GetDashboard(ctx context.Context, projectName string, dashboard
 func (c *Client) DeleteDashboard(ctx context.Context, projectName string, dashboardID string) error {
 	err := c.CallAPI(ctx, "DELETE", fmt.Sprintf("projects/%v/dashboards/%v", projectName, dashboardID), nil, nil)
 	if err != nil {
-		apiClientError := err.(APIResponseCarrier)
-		if apiClientError.GetHTTPResponse().StatusCode != http.StatusNoContent {
+		apiClientError, ok := err.(APIResponseCarrier)
+		if !ok || apiClientError.GetHTTPResponse().StatusCode != http.StatusNoContent {
```
cdion, line 118: Should this be?

`if !ok || apiClientError.GetStatusCode() != http.StatusNoContent`

heidmotron: Yes–Great catch!

PR #125, opened 2023-02-23
Update Acceptance Tests to UQL

bcronin:
This updates one of the acceptance tests from the deprecated legacy resource format to the updated UQL-based resource so that the tests pass.

Note that acceptance test run against public will not succeed until https://github.com/lightstep/lightstep/pull/40040 is deployed to public (this addresses cosmetic formatting changes to the UQL in round-tripping which could break the acceptance test).


danhurwit (approved):
LGTM

PR #126, opened 2023-02-24
Add template variable support to Terraform provider

carolynblumberg:
This PR adds template variable support to Lightstep's Terraform provider. See `docs/resources/dashboard.md` for example usage.

R: @paivagustavo @bcronin 
CC: @wfro for visibility


lightstep/resource_metric_dashboard.go
```diff
@@ -207,15 +238,44 @@ func getUnifiedDashboardAttributesFromResource(d *schema.ResourceData) (*client.
...
+			SuggestionAttributeKey: suggestionAttributeKey,
+		})
+	}
+	return newTemplateVariables
+}
+
+func buildDefaultValues(valuesIn []interface{}) []string {
+	var defaultValues []string
+	for _, v := range valuesIn {
+		defaultValues = append(defaultValues, v.(string))
```
bcronin, line 274: Is the user type, say `4` rather than `"4"` in the Terraform file as the value for the template variable, what happens? Will it end up with a panic here? 

I know the existing code doesn't set a great precedent for good user error messaging, but is there some way we can give a more informative error (assuming it does panic if a number were provided instead of a string)?

lightstep/resource_dashboard_test.go
```diff
@@ -187,6 +212,19 @@ resource "lightstep_dashboard" "test" {
 					resource.TestCheckResourceAttr(resourceName, "chart.0.query.0.dependency_map_options.0.map_type", "operation"),
 				),
 			},
+			{
```
bcronin, line 215: As someone recently debugging some of these tests, it might be nicer to split out this new test into it's own TestAccDashboardTemplateVariables test (or something similar) rather than including it as a step in a broader test.  Makes it easier to run the test in isolation if nothing else.

lightstep/resource_dashboard_test.go
```diff
@@ -131,6 +131,31 @@ resource "lightstep_dashboard" "test" {
     }
   }
 }
+`
+
+	templateVariableDashboardConfig := `
```
bcronin, line 136: Might be nice to add a configuration that's expected to fail as well.

bcronin (approved):
Looks good...

Don't forget to also bump `.go-version` though.

Also some minor comments around the tests.

lightstep/resource_metric_dashboard.go
paivagustavo: From what I remember testing earlier this week, this gets converted to `"4"` by the provider, but worth to double check :)

PR #127, opened 2023-03-01
Add dashboard labels

wfro:
Adds a new `label` option to the `lightstep_dashboard` resource. 

Labels can have a key + value or just a value, e.g.
```
resource "lightstep_dashboard" "labels" {
  project_name   = "terraform-provider-tests"
  dashboard_name = "Acceptance Test Dashboard"

  label {
    key   = "team"
    value = "ontology"
  }

  label {
    value = "terraform"
  }

  chart {
    name = "Chart Number One"
    rank = 1
    type = "timeseries"
    query {
      hidden              = false
      query_name          = "a"
      display             = "bar"
      query_string        = "metric requests | rate 10m"
    }
  }
}
```


wfro:
Still working on getting acceptance tests passing, but the changes themselves are ready for review.

bcronin (approved):

Dolphinsgrin (approved):
Looks great!

lightstep/resource_metric_dashboard.go
```diff
@@ -62,6 +62,12 @@ func resourceUnifiedDashboard(chartSchemaType ChartSchemaType) *schema.Resource
 					Schema: getChartSchema(chartSchemaType),
 				},
 			},
+			"labels": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Description: "Each label is a map with key and value fields. key is optional for free-form strings that don't use the \"key:value\" syntax.",
+				Elem:        &schema.Schema{Type: schema.TypeMap},
```
paivagustavo, line 69: Shouldn't we have a schema here as well, that way it is validated by terraform and it is also documented in the provider docs?

For example, with `TypeMap`, I think someone could write and that would be considered valid by terraform, although we do validate that later, we lose terraform own validation mechanism:
```
  labels = [
    {
      keyy = "team",
      not_really_value = "billing"
    },
  ]

Comment (wfro) [reply lightstep/resource_metric_dashboard.go:69]: @paivagustavo Yeah I agree it'd be ideal to document/validate the label object directly in terraform instead of just in code - it's an unfortunate limitation of [`TypeMap`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typemap) right now.

I originally played around with using `schema.Resource` instead of `TypeMap` which looks like

Comment (wfro) [reply lightstep/resource_metric_dashboard.go:69]: @paivagustavo Yeah I agree it'd be ideal to document/validate the label object directly in terraform instead of just in code - it's an unfortunate limitation of [`TypeMap`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-types#typemap) right now.

I originally played around with using `schema.Resource` instead of `TypeMap` which looks like
```
"label": {
	Type:        schema.TypeSet,
	Optional:    true,
	Elem: &schema.Resource{
		Schema: map[string]*schema.Schema{
			"key": {
				Optional:     true,
			},
			"value": {
				Required:     true,
			},
		},
	},
},
```

Ended up not sticking with that because it requires labels to be defined as a block vs. the list syntax that's used in a bunch of other places in the provider

Comment (paivagustavo) [reply lightstep/resource_metric_dashboard.go:69]: I see, for #126 we've been working to add template variables as a block like that:
```
  template_variable {
	name = "service"
	default_values = ["myService"]
   	suggestion_attribute_key = "service.name"
  }
  ```

I think we should align things to do look similar.

paivagustavo: Not saying that we shouldn't consider making it a list as you said, but that the provider should have a standard.

wfro: Totally agree on a standard, personal preference aside `schema.Resource` has the biggest concrete advantage in that you can take advantage of defining a schema for a map.

So proposal - update labels to use block syntax like what yall have for `template_variable`, and generally recommend that a list/set of maps should use `Resource` instead of `TypeMap` when the map structure is known ahead of time. Might not make sense to go back and change all existing usage like filters, but I think it's a good guideline going forward.

paivagustavo: Sounds like a good plan!

PR #128, opened 2023-03-01
Update ACC tests

bcronin:
* Adds several ACC tests by converting the queries to UQL and updating the resource types.
* Updates several ACC tests of legacy resource types to check for query string responses and to `ExpectNonEmptyPlan`
* Changed the query naming of one of the dashboard tests to avoid failure due to the name changing (as this is not a failure of the provider but a bug to fix elsewhere)

Streams related tests are not yet passing.


lightstep/resource_metric_condition_test.go
```diff
@@ -342,247 +294,15 @@ resource "lightstep_metric_condition" "test" {
...
 					resource.TestCheckResourceAttr(resourceName, "name", "Span latency alert - updated"),
-					resource.TestCheckResourceAttr(resourceName, "metric_query.0.spans.0.operator_input_window_ms", "3600000"),
-					resource.TestCheckResourceAttr(resourceName, "metric_query.0.spans.0.latency_percentiles.0", "95"),
-				),
-			},
-		},
-	})
-}
-
-func TestAccSpanRateCondition(t *testing.T) {
```
bcronin, line 362: This test was essentially just testing that the query round-tripped, which is redundant with other tests now that queries are all UQL.

lightstep/resource_metric_condition_test.go
```diff
@@ -342,247 +294,15 @@ resource "lightstep_metric_condition" "test" {
...
-					testAccCheckMetricConditionExists(resourceName, &condition),
-					resource.TestCheckResourceAttr(resourceName, "name", "Span rate alert - updated"),
-					resource.TestCheckResourceAttr(resourceName, "metric_query.0.spans.0.operator_input_window_ms", "3600000"),
-				),
-			},
-		},
-	})
-}
-
-func TestAccSpanErrorRatioCondition(t *testing.T) {
```
bcronin, line 477: This test was essentially just testing that the query round-tripped, which is redundant with other tests now that queries are all UQL.

lightstep/resource_metric_condition_test.go
```diff
@@ -131,15 +114,15 @@ resource "lightstep_metric_condition" "test" {
...
 
-	updatedConditionConfig := `
+	updatedConditionConfig := fmt.Sprintf(`
 resource "lightstep_slack_destination" "slack" {
   project_name = "terraform-provider-tests"
   channel = "#emergency-room"
 }
 
-resource "lightstep_metric_condition" "test" {
+resource "lightstep_alert" "test" {
```
MisterSquishy, line 125: does this belong in https://github.com/lightstep/terraform-provider-lightstep/blob/main/lightstep/resource_alert_test.go now?

lightstep/resource_metric_dashboard_test.go
```diff
@@ -15,9 +15,11 @@ import (
 func TestAccMetricDashboard(t *testing.T) {
 	var dashboard client.UnifiedDashboard
 
+	uqlQuery := `metric pagerduty.task.success | filter (kube_app == "pagerduty") | rate | group_by ["cluster-name"], max`
+
 	// missing required field 'type'
-	badDashboard := `
-resource "lightstep_metric_dashboard" "test" {
+	badDashboard := fmt.Sprintf(`
+resource "lightstep_dashboard" "test" {
```
MisterSquishy, line 22: same q, this resource has its own tests https://github.com/lightstep/terraform-provider-lightstep/blob/main/lightstep/resource_dashboard_test.go

lightstep/resource_metric_condition_test.go
```diff
@@ -216,7 +180,7 @@ resource "lightstep_metric_condition" "test" {
 					testAccCheckMetricConditionExists(resourceName, &condition),
 					resource.TestCheckResourceAttr(resourceName, "name", "Too many requests"),
 					resource.TestCheckResourceAttr(resourceName, "description", "A link to a playbook"),
-					resource.TestCheckResourceAttr(resourceName, "metric_query.0.timeseries_operator_input_window_ms", "3600000"),
+					resource.TestCheckResourceAttr(resourceName, "query.0.query_string", uqlQuery+"\n"),
```
MisterSquishy, line 183: this trailing newline is probably a bug but 🤷 a pretty inoffensive one

wfro:
Thanks for fixing these :). For the `TestAccDashboard` failure, I think changing `query_name` to "a" instead of "b" would get things passing which might be fine since it's not necessarily a bug in the provider. We should fix that (`query_name` reverting back to the default value) separately in the public API though - if I can't find a ticket I'll add one.

MisterSquishy (approved):
im in favor of merging to turn the tests green, but i wonder if the tests for legacy resource types should continue exercising the legacy format, and instead just assert that it comes back as TQL (which is how we expect those legacy resources to behave now)

lightstep/resource_metric_condition_test.go
bcronin: Hmm. I might be misinterpreting what you're saying, but the `+"\n"` is needed on all these tests since the `<<EOT` docstring syntax always adds a newline (which ends up being passed to the SaaS which does not do any trim or modification of the raw string) and the variable itself in this test code doesn't have that newline.

bcronin:
Updated the old tests to check that they are returning query strings instead of the legacy format and to `ExpectNonEmptyPlan`.

Makefile
```diff
@@ -44,6 +44,10 @@ ifndef LIGHTSTEP_API_KEY_PUBLIC
 endif
 	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_API_KEY_PUBLIC} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep
 
+tmp-test:
+	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_PUBLIC_API_KEY} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep -run TestAccSpanErrorRatioCondition
```
MisterSquishy, line 48: not a huge deal but we can dump this guy right?

PR #129, opened 2023-03-02
Do not set tql field as computed

bcronin:
Corrects the `tql` field to be not `computed`.

This fixes the failing ACC tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,9 +44,6 @@ ifndef LIGHTSTEP_API_KEY_PUBLIC
 endif
 	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_API_KEY_PUBLIC} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep
 
-tmp-test:
-	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_PUBLIC_API_KEY} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep -run TestAccSpanErrorRatioCondition
-
 
 .PHONY: ensure-clean-repo
 ensure-clean-repo:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -44,9 +44,6 @@ ifndef LIGHTSTEP_API_KEY_PUBLIC
 endif
 	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_API_KEY_PUBLIC} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep
 
-tmp-test:
-	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_PUBLIC_API_KEY} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep -run TestAccSpanErrorRatioCondition
-
```
bcronin, line 49: Debug code I accidentally included in a prior PR! (oops)

lightstep/resource_metric_condition_test.go
```diff
@@ -717,7 +717,7 @@ resource "lightstep_metric_condition" "test" {
 				Config: updatedConditionConfig,
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckMetricConditionExists(resourceName, &condition),
-					resource.TestCheckResourceAttr(resourceName, "metric_query.0.tql", "spans count | rate 1h, 1h | filter (service == \"frontend\") | group_by [], sum | point (value + value) | reduce 30s, min"),
```
bcronin, line 720: This was failing before the query string comparison so it wasn't apparent that the expected query string wasn't right either!

MisterSquishy: shouldnt we expect `value + value + value`? seems that's what the [formula update](https://github.com/lightstep/terraform-provider-lightstep/blob/03cc15b2d946c5c5de61952dfe50c8107d39dbcf/lightstep/resource_metric_condition_test.go#L685) is trying to do

MisterSquishy (approved):
worried the test is still pointing out a bug, but i think this is a good change regardless

lightstep/resource_metric_condition_test.go
MisterSquishy: (i was only reading the left hand side... 😳 this is exactly what youre fixing)

PR #130, opened 2023-03-03
Legacy query equivalence

bcronin:
This changes prevents "false diffs" in the `terraform plan` for legacy-style queries.

Internally the SaaS stores the queries as Unified Query Language (UQL) strings. This PR works by converting any legacy queries provided by Terraform to UQL form. If when GET'ing the updated resource, if the returned UQL is fully equivalent to what that legacy query, then the ResourceData for the legacy query is used rather than the ResourceData that would be used for UQL style strings.


lightstep/resource_metric_condition_test.go
```diff
@@ -216,7 +216,8 @@ resource "lightstep_metric_condition" "test" {
 					testAccCheckMetricConditionExists(resourceName, &condition),
 					resource.TestCheckResourceAttr(resourceName, "name", "Too many requests"),
 					resource.TestCheckResourceAttr(resourceName, "description", "A link to a playbook"),
-					resource.TestCheckResourceAttr(resourceName, "metric_query.0.tql", "metric requests | filter ((project_name == \"catlab\") && (service != \"android\")) | rate 1h, 1h | group_by [\"method\"], mean | reduce 30s, min"),
+					resource.TestCheckResourceAttr(resourceName, "metric_query.0.timeseries_operator_input_window_ms", "3600000"),
```
bcronin, line 219: Reverting my earlier test updates and these now _do not_ return UQL

lightstep/legacy_query_equivalence.go
```diff
@@ -0,0 +1,151 @@
...
+// checking if the the UQL translations are the same.
+func compareUpdatedLegacyQueries(
+	ctx context.Context,
+	c *client.Client,
+	projectName string,
+	priorQueries []client.MetricQueryWithAttributes,
+	updatedQueries []client.MetricQueryWithAttributes,
+) (bool, error) {
+
+	// Step 1: call the SaaS to translate the legacy queries to UQL
```
bcronin, line 91: TODO: this also needs a "step 0" of comparing all attributes of `MetricQueryWithAttributes` that _do not_ affect the query.  Or alternately, the calling code needs to be updated to only copy the resource data attributes that influence the query itself.

lightstep/legacy_query_equivalence.go
```diff
@@ -0,0 +1,151 @@
...
+// by using the legacy resource data instread of the UQL resource data.
+func dashboardHasEquivalentLegacyQueries(
+	ctx context.Context,
+	c *client.Client,
+	projectName string,
+	priorAttrs *client.UnifiedDashboardAttributes,
+	updatedAttrs *client.UnifiedDashboardAttributes,
+) (bool, error) {
+	// This code is only applicalbe if there are legacy charts
+	for _, chart := range priorAttrs.Charts {
```
MisterSquishy, line 25: do we care about catching cases where a dashboard has multiple charts where some are TQL and some are legacy? i think this will short circuit if any chart on the dashboard has a tql query

lightstep/legacy_query_equivalence.go
MisterSquishy: slight bias towards the first one; seems clearer than this function requiring some special pre-work in order to be safely callable

.go-version
```diff
@@ -1 +1 @@
-1.71.1
+1.72.0
```
MisterSquishy, line 1: nit: patch as this is technically "just" a bugfix?

MisterSquishy (approved):
🥳

PR #140, opened 2023-03-21
Add support for hidden_queries

bcronin:
Adds Terraform support for the `hidden_queries` field from the dashboard public API.

This allows the Terraform user to specify the hidden state of subqueries within a query string when the query is visualized.

Note that due to a limitation of Terraform, the hidden state must be specified with boolean _strings_ of `"true"` and `"false"` (not simply `true` and `false`). Also, due to how the public API works, the user must specify the state for each of the subqueries if this optional field is used.

Example:

```hcl
resource "lightstep_dashboard" "my_dashboard" {
	project_name          = "my_project"
	dashboard_name        = "My Dashboard"
	chart {
		name = "Chart Number One"
		rank = 1
		type = "timeseries"
		query {
			hidden              = false
			query_name          = "a"
			display             = "line"
			query_string        = <<EOT
with
	b = metric m | filter (service == "cats") | rate | group_by [], sum;
	c = metric m | filter (service == "dogs") | rate | group_by [], sum;
join b + c, b = 0, c = 0
EOT

			hidden_queries = {
				b = "false"
				c = "true"
			}
		}
	}
}
```


carolynblumberg (approved):
Looks great! Sorry for the delay, was playing around with it. The string boolean requirement is funky, but looks like it'll be converted correctly even if a user types `true` instead of `"true"` 👍

PR #143, opened 2023-04-14
[Composite Alerts] prepare Expression to become Optional at times

mattcarmody:
This is part 1 of adding Composite Alerts to the terraform provider. This should be a no-op, it just lays the foundation for upcoming changes.

### DIFF
--- a/client/metric_conditions.go
+++ b/client/metric_conditions.go
@@ -18,7 +18,7 @@ type UnifiedConditionAttributes struct {
 	Description   string                      `json:"description"`
 	Labels        []Label                     `json:"labels"`
 	Type          string                      `json:"condition_type"`
-	Expression    *Expression                 `json:"expression,omitempty"`
+	Expression    Expression                  `json:"expression"`
 	Queries       []MetricQueryWithAttributes `json:"metric-queries"`
 	AlertingRules []AlertingRule              `json:"alerting-rules,omitempty"`
 }
--- a/client/metric_conditions.go
+++ b/client/metric_conditions.go
@@ -30,9 +30,13 @@ type AlertingRule struct {
 }
 
 type Expression struct {
+	SubAlertExpression
+	IsMulti bool `json:"is-multi-alert,omitempty"`
+}
+
+type SubAlertExpression struct {
 	Thresholds Thresholds `json:"thresholds"`
 	Operand    string     `json:"operand"`
-	IsMulti    bool       `json:"is-multi-alert,omitempty"`
 	IsNoData   bool       `json:"enable-no-data-alert,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


lightstep/resource_dashboard.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-func getUnifiedQuerySchema() map[string]*schema.Schema {
+func getUnifiedQuerySchemaMap() map[string]*schema.Schema {
```
mattcarmody, line 12: a bunch of functions are named with the same pattern. some return a schema while others return a map of schemas so I renamed some

client/metric_conditions.go
```diff
@@ -30,9 +30,13 @@ type AlertingRule struct {
 }
 
 type Expression struct {
+	SubAlertExpression
```
mattcarmody, line 33: `SubAlertExpression` will be used without `IsMulti` in follow up PR(s)

lightstep/resource_metric_condition.go
```diff
@@ -344,6 +312,66 @@ func getThresholdSchema() map[string]*schema.Schema {
...
+	resource := getCompositeSubAlertExpressionResource()
+	resource.Schema["is_multi"] = getIsMultiSchema()
+	return &schema.Schema{
+		Type:     schema.TypeList,
+		Required: true,
+		MaxItems: 1,
+		MinItems: 1,
+		Elem:     resource,
+	}
+}
```
mattcarmody, line 338: these are the separate paths of the branch above and they're quite similar, the difference is Required/Optional & MinItems. I think breaking things out this way will make a bit more sense when I add the `SubAlert` version

lightstep/resource_metric_condition.go
```diff
@@ -488,9 +516,7 @@ func (p *resourceUnifiedConditionImp) resourceUnifiedConditionImport(ctx context
 }
 
 func getUnifiedConditionAttributesFromResource(d *schema.ResourceData, schemaType ConditionSchemaType) (*client.UnifiedConditionAttributes, error) {
-	expression := d.Get("expression").([]interface{})[0].(map[string]interface{})
-
-	thresholds, err := buildThresholds(d)
```
mattcarmody, line 493: `buildThresholds` makes assumptions about `expression` being present and ends up in the `Expression` so I moved it inside `buildExpression`

gontzess (approved):
looks good, LGTM!

lightstep/resource_metric_condition.go
```diff
@@ -344,6 +312,67 @@ func getThresholdSchema() map[string]*schema.Schema {
 	}
 }
 
+// expression is optional because it cannot be included in composite alerts but is otherwise required
```
lambcode, line 315: I'm not sure I follow this comment. There doesn't appear to be anything marked optional here.

lambcode (commented):
Sorry for the late review, but I have a few questions on this one.

PR #145, opened 2023-04-21
fix legacy implicit groups detection

MisterSquishy:
fixes a bug in which the legacy implicit group detector was incorrectly returning true for _explicit_ implicit groups.

let's assume we have a dashboard with no explicit groups, and no fancy chart positions. we ought to represent that in HCL like so:
```
resource "lightstep_dashboard" "foo" {
...
group {
    rank          = 0
    title           = ""
    visibility_type = "implicit"
    ...charts...
}
```

this representation will incorrectly be identified as a "legacy" implicit group, which was a term we made up to describe HCL like the below:
```
resource "lightstep_dashboard" "foo" {
...
...charts...
```

for HCL with "legacy" implicit groups, we persist the loose charts in state (so as not to show a diff with the existing state). but for HCL like the first example, we do _not_ want to persist the loose charts in state, since there is an explicit group. if we do so, then every subsequent plan will show a diff (deletion of the loose charts).

this PR fixes that by assigning a unique title to the synthetically-constructed "legacy" implicit groups, so they can be distinguished from the _explicit_ implicit groups.


lightstep/resource_metric_dashboard.go
```diff
@@ -18,6 +18,10 @@ type ChartSchemaType int
 const (
 	MetricChartSchema ChartSchemaType = iota
 	UnifiedChartSchema
+
+	// if the incoming HCL contains loose charts, we will put them in an implicit group
+	// this group should have a unique name, so we can distinguish it from an _explicit_ implicit group with no title
+	legacyImplicitGroupName = "legacy implicit group"
```
MisterSquishy, line 24: yes, someone might explicitly name their implicit group`"legacy implicit group"`, which would break this and result in a spurious diff. this seems exceedingly unlikely. why would anyone name their implicit group anyway.

danhurwit: By definition, if an implicit group is named, it is an explicit group

danhurwit (approved):
LGTM. Want to make sure that the implicit group title does not get displayed in the FE if we persist it.

lightstep/resource_metric_dashboard.go
```diff
@@ -649,13 +651,16 @@ func (p *resourceUnifiedDashboardImp) setResourceDataFromUnifiedDashboard(projec
 
 // isLegacyImplicitGroup defines the logic for determining if the charts in this dashboard need to be unwrapped to
 // maintain backwards compatibility with the pre group definition
-func isLegacyImplicitGroup(groups []client.UnifiedGroup) bool {
+func isLegacyImplicitGroup(groups []client.UnifiedGroup, hasLegacyChartsIn bool) bool {
 	if len(groups) != 1 {
```
danhurwit, line 655: Do we need these other checks here anymore?

PR #150, opened 2023-05-11
use default retryablehttp client

paivagustavo:
The current client does not retry 429 errors, which can led to failed plan/apply that could be resolved with a few retries in those cases.

This changes to use the default `retryablehttp.Client` which retries all 5xx and 429 erros, up to 5 times.


MisterSquishy (approved):
🥳

PR #151, opened 2023-05-11
Improve example lightstep_alert config to include alerting rule

lambcode:
Not having the alerting rule in the example has caused some confusion. Especially given there is a alerting_rule resource that is old and shouldn't be used anymore. Also added notes on the resources that will be deprecated.


docs/resources/alert.md
```diff
@@ -39,6 +39,11 @@ spans latency
 | point percentile(value, 50)
 EOT
   }
+
+  alerting_rule {
+      id = lightstep_pagerduty_destination.my_destination.id
+  }
```
mattcarmody, line 45: would it be helpful to readers to include a frequency?

docs/resources/stream_condition.md
```diff
@@ -6,6 +6,8 @@ description: |-
 
 ---
 
+# NOTE: this resource will be deprecated in v2+. Use `lightstep_alert` instead.
```
mattcarmody, line 9: while in here marking things for future deprecation, is `metric_condition` another candidate to be deprecated?

mattcarmody (approved):
🎉 love helpful docs

docs/resources/stream_condition.md
lambcode: `metric_condition` already has a deprecation note on it :)

PR #154, opened 2023-05-16
Add servicenow destination resource to provider

lambcode:
Adds the upcoming ServiceNow destination type to the lightstep terraform provider


lightstep/resource_servicenow_destination.go
```diff
@@ -0,0 +1,129 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	"github.com/lightstep/terraform-provider-lightstep/client"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
```
mattcarmody, line 13: nit: this ordering looks off?

client/destinations.go
```diff
@@ -32,6 +32,19 @@ type SlackAttributes struct {
...
+	Name            string `json:"name"`
+	DestinationType string `json:"destination_type"`
+	URL             string `json:"url"`
+	Auth            Auth   `json:"auth"`
+}
+
+type Auth struct {
+	Username string `json:"username"`
+	// Password is only set for requests. Will be empty for API responses
+	Password string `json:"password"`
```
mattcarmody, line 45: if we don't return this field in responses I would expect the diff to complain all the time, what makes this ok?

mattcarmody (approved):
🎉

client/destinations.go
lambcode: Terraform will use the value saved in its state file to compare to the value in the *.tf file if the field isn't returned by the GET api. I manually tested that this works by applying a terraform file with one of these servicenow destinations then running `terraform plan` and seeing that there are no changes detected.

PR #155, opened 2023-05-17
Text Panel Updates

bcronin:
Adds `text_panel` support.


client/metric_dashboards.go
```diff
@@ -47,6 +47,7 @@ type UnifiedChart struct {
 	ChartType     string                      `json:"chart-type"`
 	YAxis         *YAxis                      `json:"y-axis"`
 	MetricQueries []MetricQueryWithAttributes `json:"metric-queries"`
+	Text          string                      `json:"text"`
```
bcronin, line 50: Add the `text` field to mirror the public API request struct.

lightstep/resource_metric_dashboard.go
```diff
@@ -126,46 +126,37 @@ func getGroupSchema(chartSchemaType ChartSchemaType) map[string]*schema.Schema {
...
-
-	var querySchema map[string]*schema.Schema
-	if chartSchemaType == UnifiedChartSchema {
-		querySchema = getUnifiedQuerySchemaMap()
-		querySchema["dependency_map_options"] = &schema.Schema{
+		"text_panel": {
 			Type:     schema.TypeList,
 			Optional: true,
-			MaxItems: 1,
+			Computed: true, // the panels can be mutated individually; chart mutations should not trigger group updates
```
bcronin, line 132: Following the pattern we used for "charts" and setting `Computed` to `true`, see https://github.com/lightstep/terraform-provider-lightstep/pull/152

lightstep/resource_metric_dashboard.go
```diff
@@ -126,46 +126,37 @@ func getGroupSchema(chartSchemaType ChartSchemaType) map[string]*schema.Schema {
...
+	return mergeSchemas(
+		getPanelSchema(),
+		map[string]*schema.Schema{
+			"text": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
+	)
+}
```
bcronin, line 150: Split out the common chart + text panel schema elements into `getPanelSchema` so `"text"` is the only unique field to text panels.

lightstep/resource_metric_dashboard.go
```diff
@@ -126,46 +126,37 @@ func getGroupSchema(chartSchemaType ChartSchemaType) map[string]*schema.Schema {
...
 		"name": {
 			Type:     schema.TypeString,
-			Required: true,
-		},
-		"rank": {
-			Type:         schema.TypeInt,
-			ValidateFunc: validation.IntAtLeast(0),
-			Required:     true,
+			Optional: true,
+			Default:  "",
```
bcronin, line 159: Made "name" (aka "Title") optional. It is should be optional for text panels and on the API/product level charts don't require a title either.

lightstep/resource_metric_dashboard.go
```diff
@@ -191,41 +182,80 @@ func getChartSchema(chartSchemaType ChartSchemaType) map[string]*schema.Schema {
...
 			Computed: true,
 		},
-		"y_axis": {
-			Type:       schema.TypeList,
-			MaxItems:   1,
-			Deprecated: "The y_axis field is no longer used",
+	}
+}
+
+func getChartSchema(chartSchemaType ChartSchemaType) map[string]*schema.Schema {
```
bcronin, line 192: `getChartSchema` is unchanged other than moved code and splitting out the common schema to a call to `getPanelSchema`.

lightstep/resource_metric_dashboard.go
bcronin: Also "text_panel" uses `TypeList` (which retains the order) rather than `TypeSet`. I'm not aware of why `TypeSet` was used for charts, but it seems only to make things slightly more complex since order is not retained.  (If I'm missing anything here, this is easy to change.)

lightstep/resource_metric_dashboard.go
```diff
@@ -430,27 +459,56 @@ func buildGroups(groupsIn []interface{}, legacyChartsIn []interface{}) ([]client
...
 }
 
+func buildPosition(panel map[string]interface{}) client.UnifiedPosition {
+	return client.UnifiedPosition{
+		XPos:   panel["x_pos"].(int),
+		YPos:   panel["y_pos"].(int),
+		Width:  panel["width"].(int),
+		Height: panel["height"].(int),
+	}
+}
```
bcronin, line 493: Split out into a helper now that it's called in multiple places.

lightstep/resource_metric_dashboard.go
```diff
@@ -561,50 +613,14 @@ func (p *resourceUnifiedDashboardImp) setResourceDataFromUnifiedDashboard(projec
...
-		return charts, nil
-	}
 	if isLegacyImplicitGroup(dash.Attributes.Groups, hasLegacyChartsIn) {
-		charts, err := assembleCharts(dash.Attributes.Groups[0].Charts)
+		charts, textPanels, err := assembleDashboardPanels(dash.ID, p.chartSchemaType, dash.Attributes.Groups[0].Charts)
 		if err != nil {
 			return err
 		}
+		if len(textPanels) > 0 {
+			return fmt.Errorf("text panels are only supported within groups")
```
bcronin, line 622: AFAIK, we don't intend to support text panels outside of groups as everything is intended to be in groups moving forward.

lightstep/resource_metric_dashboard.go
```diff
@@ -650,6 +668,125 @@ func (p *resourceUnifiedDashboardImp) setResourceDataFromUnifiedDashboard(projec
...
+	chartResources, err = assembleCharts(dashboardID, chartSchemaType, charts)
+	if err != nil {
+		return nil, nil, err
+	}
+	textPanelResources, err = assembleTextPanels(dashboardID, textPanels)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return chartResources, textPanelResources, nil
```
bcronin, line 707: (Adding the return value names to be explicit / more clear though I know they're not strictly required)

lightstep/resource_metric_dashboard.go
```diff
@@ -650,6 +668,125 @@ func (p *resourceUnifiedDashboardImp) setResourceDataFromUnifiedDashboard(projec
...
+	chartsIn []client.UnifiedChart,
+) ([]interface{}, error) {
+	var chartResources []interface{}
+	for _, c := range chartsIn {
+		if c.ChartType != "timeseries" {
+			return nil, fmt.Errorf("panel %s is not a timeseries chart", c.Title)
+		}
+
+		resource := map[string]interface{}{}
+		setPanelResourceData(resource, c)
```
bcronin, line 758: Moving the common code to `setPanelResourceData` is the only change to `assembleCharts`, otherwise it's just moved code.

lightstep/resource_metric_dashboard.go
danhurwit: I would stay away from these as they add the potential for unintended sideffectyness.

danhurwit (approved):
LGTM

lightstep/util.go
```diff
@@ -0,0 +1,13 @@
+package lightstep
+
+import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+func mergeSchemas(arr ...map[string]*schema.Schema) map[string]*schema.Schema {
+	dst := make(map[string]*schema.Schema)
```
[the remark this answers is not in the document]
bcronin: I don't really think that's necessary. I don't think "last write wins" is that surprising for a utility function like this.  🤷 if you feel strongly, I'm glad to update it.

PR #157, opened 2023-05-31
remove query_format argument

MisterSquishy:
this argument is no longer needed, as all queries are persisted in their query string format


MisterSquishy:
not versioning since there is no behavior change, this is purely cleaning up unneeded code (i think that's kosher?)

bcronin (approved):

PR #158, opened 2023-06-14
change acceptance test project and bump go version

paivagustavo:
* updates to go `1.20`
* updates tfplugindocs to `v0.15.0`
* change the lightstep project used for acceptance tests
* make `name` required for charts (text panels remains optional)


Makefile
```diff
@@ -42,7 +42,7 @@ acc-test:
 ifndef LIGHTSTEP_API_KEY_PUBLIC
 	$(error LIGHTSTEP_API_KEY_PUBLIC must be defined for acc-test)
 endif
-	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_API_KEY_PUBLIC} LIGHTSTEP_ORG="LightStep" LIGHTSTEP_ENV="public" go test -v ./lightstep
+	@TF_ACC=true LIGHTSTEP_API_KEY=${LIGHTSTEP_API_KEY_PUBLIC} LIGHTSTEP_ORG="terraform-provider" LIGHTSTEP_PROJECT="terraform-provider-test" LIGHTSTEP_ENV="public" go test -v ./lightstep
```
bcronin, line 45: Do the `test-local` and `test-staging` targets also need to be updated?

lightstep/resource_alert_test.go
```diff
@@ -17,7 +17,7 @@ func TestAccAlert(t *testing.T) {
 
 	badAlertMissingQueryAndCompositeFields := `
 resource "lightstep_alert" "errors" {
-  project_name = "terraform-provider-tests"
+  project_name = ` + fmt.Sprintf("\"%s\"", testProject) + `
```
bcronin, line 20: I know this is just a test, but given this is repeated many times, would it make sense to simplify all these `fmt.Sprintf` calls...

From:

```go
project_name = ` + fmt.Sprintf("\"%s\"", testProject) + `
```

To:

```go
project_name = "` + testProject + `"
```

...or maybe move it to a common variable in some form?

lightstep/resource_metric_dashboard.go
```diff
@@ -150,14 +150,27 @@ func getTextPanelSchema() map[string]*schema.Schema {
...
+			}
+		}
+
+		return &schema.Schema{
 			Type:     schema.TypeString,
 			Optional: true,
 			Default:  "",
-		},
+		}
+	}
```
bcronin, line 169: Nit: is a closure needed here?  Would it be more straightforward to put the logic inline?

```go
	return map[string]*schema.Schema{
		// Alias for what we refer to as title elsewhere
		"name": &schema.Schema{
			Type:     schema.TypeString,
			Required: isNameRequired,
            Optional: !isNamedRequired,
			Default:  "",
        },
        ...
```

bcronin (approved):

lightstep/resource_metric_dashboard.go
paivagustavo: Yeah, I tried that, but if `Optional: true` we can't set anything to `Default`.

lightstep/resource_alert_test.go
paivagustavo: Yup, this was me being lazy because this was already done in some stream alerts, so I just copied what was done there, but update it with your suggestion!

Makefile
paivagustavo: For some reason I thought this was going to be okay for staging but it is definitely better to have this separate as well.

PR #166, opened 2023-08-07
Mark the threshold-related fields as optional in the unified alert resource schema

ltyson:
In the Lightstep UI and API it is valid to configure an alert with just the "No data" option selected (no Critical/Warning thresholds), however the terraform provider currently rejects alerts with that configuration. This PR changes that logic to treat all fields in the `Expression` block as optional, relying instead on the Lightstep API handler to validate that the alert contains at least one of:
* the No Data option selected
* at least one of the Critical / Warning thresholds configured

R: @lambcode @mattcarmody


mattcarmody (approved):

ltyson:
It turns out I did this correctly, but the public API currently also rejects requests of this shape so I need to merge a different PR first before these tests will pass.

lambcode:
Did you see [this comment ](https://lightstep.atlassian.net/browse/LS-51301?focusedCommentId=67529) on the jira. I think its worth spending some time to ensure that these edge cases export properly from the UI terraform export button in a way that actually works in the terraform provider.

lightstep/resource_metric_condition.go
```diff
@@ -966,7 +982,15 @@ func buildKeys(keysIn []interface{}) []string {
 func buildThresholds(singleExpression map[string]interface{}) (client.Thresholds, error) {
 	t := client.Thresholds{}
 
-	thresholdsObj := singleExpression["thresholds"].([]interface{})[0].(map[string]interface{})
+	elem, ok := singleExpression["thresholds"]
```
ltyson, line 985: previously `thresholds` was marked as required so we could count on there being a non-empty list, but now it's optional so we can't

lightstep/resource_alert_test.go
```diff
@@ -680,22 +680,93 @@ resource "lightstep_alert" "errors" {
 `
 
 	compositeConditionConfig := `
+	resource "lightstep_slack_destination" "slack" {
```
ltyson, line 683: the diff is a little weird here -- I didn't change the existing test alerts, just added a few new ones that are largely identical (so github got confused)

lightstep/resource_metric_condition.go
```diff
@@ -322,12 +322,10 @@ func getThresholdSchemaMap() map[string]*schema.Schema {
 		"critical": {
 			Type:     schema.TypeString,
 			Optional: true,
-			Default:  "",
```
ltyson, line 325: these default values were causing a non-nil thresholds block to be added even when it's empty so I removed them -- they're both optional so that seems fine to me

lightstep/resource_metric_condition.go
```diff
@@ -395,21 +393,39 @@ func getCompositeSubAlertExpressionSchema() *schema.Schema {
...
 			"thresholds": {
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					if strings.HasSuffix(k, "thresholds.#") {
+						// When parsing the resource that we get from the Lightstep API, we don't include the thresholds
+						// block unless at least one of the thresholds is set. If a user includes an empty thresholds
+						// block in their config instead of omitting it entirely, terraform would normally treat that as a diff.
+						// However, for convenience, we allow users to specify an empty thresholds block and
```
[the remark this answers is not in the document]
ltyson: That was the plan, but the validation function I was going to use to return a nice error message in that scenario doesn't work for List or Set types so it turned out to be easier to just support both than to return an error message for one of them 🤦‍♀️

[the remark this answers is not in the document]
lambcode: I see. Thanks for explaining! SGTM

PR #171, opened 2023-08-17
Add scale and log base to display type optons

srjames90:
Updates the display type options to have y-scale and log base


srjames90:
I added some tests, but the DisplayTypeOptioins test always fails for me when running the tests as indicated in the readme regardless if I make changes or not. When I try running them in vscode, it seems to pass  no matter what, so will add them ni a following commit.

lightstep/resource_dashboard.go
```diff
@@ -135,6 +143,9 @@ func displayTypeOptionsFromResourceData(opts map[string]interface{}) *schema.Set
 	f := func(i interface{}) int {
 		return 1
 	}
+	if opts["y_axis_scale"] == "log" && opts["y_axis_log_base"] == "" {
+		opts["y_axis_log_base"] = 10
```
lambcode, line 147: Won't this make the terraform diff always show a difference when the `y_axis_log_base` is not set? It will always be comparing `10` to the empty string, right?

srjames90: Do you mean the diff on apply? I'm seeing it on every apply anyway. Unless you're referring to something else. I'm a noob at using terraform. I could also try figuring out where it's being saved and do it there (I'm guessing the public AP)
Screenshot 2023-08-17 at 5 49 44 PM

lambcode: yeah the diff on the apply. I suspect that it would always see a change because it compares what is in the terraform file with what comes back from the GET endpoint (that we then mutate here)

lambcode: What I'd expect is the the first time it's applied there is a diff, but after that there is no diff

srjames90: Yeah, oddly, I see a diff for display_type_options happening even without my changes.

PR #172, opened 2023-08-21
LS-51876 Make inferred_service_rule resource acceptance tests more resilient

gdfast:
What does this PR do?
---------------------

Adds a random five-character string suffix to each inferred service rule name in the inferred_service_rule resource acceptance test

For what purpose?
---------------------

Test failures can occasionally leave inferred service rules in the target Lightstep project that are not cleaned up. Subsequent test runs can fail due to rule name collisions because internal logic requires unique rule names. By prepending a random string suffix, previous test failures are no longer likely to cause another test failure

Test Plan
---------------------

- This is an update to the acceptance tests, so the acceptance tests continuing to pass is the most essential check
- I stopped the test mid-execution with a debugger to prove that two inferred service rules were created in Lightstep with random string suffixes in the names

### DIFF
--- a/lightstep/resource_inferred_service_rule_test.go
+++ b/lightstep/resource_inferred_service_rule_test.go
@@ -3,18 +3,23 @@ package lightstep
 import (
 	"context"
 	"fmt"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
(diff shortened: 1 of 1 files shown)


carolynblumberg (approved):
👍

PR #174, opened 2023-08-24
LS-45128: Support custom data field for unified alerts

ltyson:
What does this PR do?
----------------------
Adds support for the `custom-data` field in the LS terraform provider (we forgot to do this initially when adding it to the public API)

R: @lightstep/team-alerts-be

### DIFF
--- a/lightstep/resource_metric_condition_test.go
+++ b/lightstep/resource_metric_condition_test.go
@@ -157,6 +157,8 @@ resource "lightstep_metric_condition" "test" {
 	  }
   }
 
+  custom_data = "this string could be json { example_key: \"example value\"} or anything you want"
+
   metric_query {
     metric         = "requests"
     query_name          = "a"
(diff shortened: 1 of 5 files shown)


lightstep/resource_metric_condition_test.go
```diff
@@ -331,7 +333,8 @@ resource "lightstep_metric_condition" "test" {
 					testAccCheckMetricConditionExists(resourceName, &condition),
 					resource.TestCheckResourceAttr(resourceName, "name", "Too many requests"),
 					resource.TestCheckResourceAttr(resourceName, "description", "A link to a playbook"),
-					resource.TestCheckNoResourceAttr(resourceName, "labels"),
```
lambcode, line 334: Any idea why we checked this before?

lambcode (approved):
LGTM

lightstep/resource_metric_condition_test.go
ltyson: I think the idea was to make sure no labels had been set, but it wasn't working properly because the name of the field in the schema is `label` rather than `labels` so this check was misleading (it wasn't doing what it was supposed to be doing). I just changed it to check the thing it was supposed to check.

ltyson: I noticed it because I was trying to model my new check after the labels one and couldn't figure out why mine was failing while this was succeeding, then realized why.

PR #176, opened 2023-08-28
Add opsgenie destination resource

lambcode:
Adds a resource to manage opsgenie destinations.


lightstep/resource_opsgenie_destination.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+			"auth": {
+				Type:        schema.TypeList,
+				MinItems:    1,
+				MaxItems:    1,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Basic auth used to authenticate with the Opsgenie instance",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
```
ltyson, line 50: could you add descriptions for these fields too? it's on my mind since I just went through and did it for a bunch of things

lightstep/resource_opsgenie_destination.go
```diff
@@ -0,0 +1,128 @@
...
+						"username": {
```
ltyson, line 51: I thought the username wasn't used with OpsGenie destination?

lightstep/provider_test.go
```diff
@@ -24,7 +24,7 @@ func init() {
 
 	testProject = os.Getenv("LIGHTSTEP_PROJECT")
 	if testProject == "" {
-		testProject = "terraform-provider-tests"
+		testProject = "terraform-provider-test"
```
ltyson, line 27: thank you! I didn't realize this was already set automatically, I had just made an env variable

lightstep/resource_opsgenie_destination.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+
+	"github.com/lightstep/terraform-provider-lightstep/client"
+)
+
+func resourceOpsgenieDestination() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: resourceOpsgenieDestinationCreate,
```
ltyson, line 17: do you want to add a note for future readers about why there's no Update entry here?

lightstep/resource_opsgenie_destination.go
```diff
@@ -0,0 +1,128 @@
...
+	destination, err := c.CreateDestination(ctx, d.Get("project_name").(string), dest)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("failed to create Opsgenie destination %v: %v", attrs.Name, err))
+	}
+
+	d.SetId(destination.ID)
+	return resourceDestinationRead(ctx, d, m)
+}
+
+func resourceOpsgenieDestinationImport(ctx context.Context, d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
```
ltyson, line 95: ooc, how do we test this sort of thing?

lightstep/resource_opsgenie_destination_test.go
```diff
@@ -0,0 +1,150 @@
...
+				),
+				ExpectError: regexp.MustCompile("Insufficient auth blocks"),
+			},
+			{
+				Config: destinationConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckWebhookDestinationExists("lightstep_opsgenie_destination.opsgenie", &destination),
+					resource.TestCheckResourceAttr("lightstep_opsgenie_destination.opsgenie", "destination_name", "my-destination"),
+					resource.TestCheckResourceAttr("lightstep_opsgenie_destination.opsgenie", "url", "https://example.com"),
+					resource.TestCheckResourceAttr("lightstep_opsgenie_destination.opsgenie", "auth.0.username", ""),
```
ltyson, line 72: what happens if they do specify a username? is that an option we even want to expose if it will never be needed?

lightstep/resource_opsgenie_destination_test.go
```diff
@@ -0,0 +1,150 @@
...
+  project_name = ` + fmt.Sprintf("\"%s\"", testProject) + `
+  destination_name = "my-destination"
+  url = "https://example.com"
+}
+`
+
+	destinationConfig := `
+resource "lightstep_opsgenie_destination" "opsgenie" {
+  project_name = ` + fmt.Sprintf("\"%s\"", testProject) + `
+  destination_name = "my-destination"
```
ltyson, line 39: nit: just to make this more realistic, maybe add some spaces in the name?

lightstep/resource_opsgenie_destination_test.go
lambcode: I waffled on this. On one hand opsgenie is accepting a basic auth token so it makes sense to support username if they ever start using it. On the other hand we could hide this from the user in terraform just like we do in the UI. (It wouldn't be hidden in the public API though.

lambcode: Do you have a preference?

ltyson: I'd vote to hide it in terraform, assuming that doesn't create more work for you, in the interest of making terraform as minimally confusing as possible.

ltyson (approved):
I left some comments but nothing blocking, lgtm

PR #177, opened 2023-08-28
Remove references to match-on filters other than equality (and more docs updates)

ltyson:
What does this PR do and why?
----------------------
Removes references to `filters` and `exclude_filters` from the alerting_rules field in a unified alert configuration, because those filter types don't actually work in this context and never should have been added in the first place. Technically this is a backward-incompatible change, but I think this is okay because (1) I verified that none of these filter types are being used in production today and (2) the docs were previously nonexistent so I don't see how anyone would have successfully configured them anyway.

I also made a few more small docs updates while I was in the area.


templates/resources/alert.md.tmpl
```diff
@@ -1,5 +1,10 @@
+<!---
```
ltyson, line 1: I guess technically these docs only said they were generated (initially), not auto-generated (every time), but that distinction was subtle and just generally confusing so I removed it in favor of this. Same with other files below.

docs/resources/alert.md
```diff
@@ -1,5 +1,10 @@
+<!---
```
ltyson, line 1: ugh, this isn't supposed to end up here. I need to figure out a way to suppress this when these files are auto-generated from the .md.tmpl files. Maybe I'll just remove this whole block for now to avoid confusing people even more.

docs/resources/alerting_rule.md
```diff
@@ -1,5 +1,4 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
```
ltyson, line 2: I removed this line because it kept tripping me up thinking that these files were generated _every time_ as opposed to the reality that they were initially generated as convenient placeholders but from that point onward they are manually edited by us.

mattcarmody: are these files not updated by running the tfplugindocs command the DEVELOPMENT.md file describes?

mattcarmody (approved):
LGTM! 🪓

PR #178, opened 2023-08-29
Add a blurb for renaming of Lightstep to ServiceNow Cloud Observability

neilwa:
Add a blurb for renaming of Lightstep to ServiceNow Cloud Observability


jknostman29 (approved):
Looks good to me

PR #181, opened 2023-09-07
add y axis display options

MisterSquishy:
adds display option values for `y_axis_scale`, `y_axis_log_base`, `y_axis_min` and `y_axis_max`. the former two have some basic enum-style validation, the latter two have none. NB i chose _not_ to validate things like "log base is only set when y axis scale is log" or "max is greater than min" because this felt like a deep rabbit hole with potential to accidentally prevent valid configurations


lightstep/resource_dashboard.go
```diff
@@ -48,6 +48,31 @@ func getUnifiedQuerySchemaMap() map[string]*schema.Schema {
...
+					"y_axis_log_base": {
+						Type:     schema.TypeInt,
+						Optional: true,
+						ValidateFunc: validation.IntInSlice([]int{
+							2,
+							10,
+						}),
+					},
+					"y_axis_min": {
+						Type:     schema.TypeFloat,
```
MisterSquishy, line 69: float matches what we do for the [deprecated version](https://github.com/lightstep/terraform-provider-lightstep/blob/c6e0c48de3cfb56ae66d8658cbb6a779ae9581e5/lightstep/resource_metric_dashboard.go#L257) of these options, i assume we do this so these values can get very large

bcronin (approved):

PR #185, opened 2023-10-11
Fix issue updating a composite alert causes a panic in the provider

adcharre:
## Description
Suppressing the diff results in an empty "alert" element existing during the planning stage. This resulting in a panic, when applying, in `buildCompositeAlert()` line 1090 as "expressions" doesn't exist in the empty "alert" element.

This issue was introduced in v1.82.0 when the diff suppression function was added to the thresholds block.

## Steps to reproduce:

Using the attached example terraform, deploy the alert. Once deployed, modified one of the queries in the composite alert in alert.tf and plan the changes.

You should see something like the following, I've annotated (👈 ) the output below to highlight the problem
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # lightstep_alert.test_bug will be updated in-place
  ~ resource "lightstep_alert" "test_bug" {
        id           = "m9wZ5RZCN"
        name         = "TestCompositeAlertBug"
        # (2 unchanged attributes hidden)

      ~ composite_alert {
          - alert {
              - name = "A" -> null

              - expression {
                  - is_no_data = false -> null
                  - operand    = "above" -> null

                  - thresholds {
                      - warning = "95" -> null
                    }
                }

              - query {
                  - hidden         = false -> null
                  - hidden_queries = {} -> null
                  - query_name     = "a" -> null
                  - query_string   = <<-EOT
                        with
                        m2=metric tstv_capture_notifications_received_count|filter eventType=~"Complete|Partial"|rate 30m,30s|group_by [],sum;
                        m3=with
                        m0=metric tstv_capture_notifications_received_count|filter eventType=~"Complete|Partial|CaptureError"|rate 30m,30s|group_by [],sum;
                        m1=metric tstv_capture_notifications_received_count|filter eventType=="Failed"|rate 30m,30s|point value / 4|group_by [],sum;
                        join m0 + m1,m1=0;
                        join m2 / m3|point value * 100|reduce 30m,min
                    EOT -> null
                }
            }
          - alert {
              - name = "B" -> null

              - expression {
                  - is_no_data = false -> null
                  - operand    = "below" -> null

                  - thresholds {
                      - warning = "97.000001" -> null
                    }
                }

              - query {
                  - hidden         = false -> null
                  - hidden_queries = {} -> null
                  - query_name     = "b" -> null
                  - query_string   = <<-EOT
                        with
[...]

PR #186, opened 2023-10-17
bug fix + tests

ltyson:
Incorporates this [original fix](https://github.com/lightstep/terraform-provider-lightstep/pull/185) from @adcharre (which eloquently describes the bug), adds a regression test for it, and bumps the provider version.

### DIFF
--- a/examples/versions.tf
+++ b/examples/versions.tf
@@ -7,7 +7,7 @@ terraform {
   required_providers {
     lightstep = {
       source  = "lightstep/lightstep"
-      version = "~> 1.61.1"
+      version = "~> 1.61.2"
     }
   }
   required_version = "~> 1.1.0"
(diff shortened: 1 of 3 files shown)


lightstep/resource_alert_test.go
```diff
@@ -123,14 +123,14 @@ resource "lightstep_alert" "test" {
 			{
 				Config: badAlertMissingQueryAndCompositeFields,
 				Check: resource.ComposeTestCheckFunc(
-					testAccChecLightstepAlertExists(resourceName, &condition),
+					testAccCheckLightstepAlertExists(resourceName, &condition),
```
ltyson, line 126: there was a typo in the name of this function and it was bugging me :)

mattcarmody (approved):
LGTM 🎉

examples/versions.tf
```diff
@@ -7,7 +7,7 @@ terraform {
   required_providers {
     lightstep = {
       source  = "lightstep/lightstep"
-      version = "~> 1.61.1"
+      version = "~> 1.61.2"
```
ltyson, line 10: oops, I changed the wrong file to update the version number 🤦‍♀️ this was totally not it

ltyson: Here's the follow-up: https://github.com/lightstep/terraform-provider-lightstep/pull/190

PR #193, opened 2023-10-27
add validation on query name length

MisterSquishy:
the maximum allowed length of a query name is 128 characters, we should enforce that in the provider


bcronin (approved):
👍 

(though wish we didn't now have 128 written in 3 places and _not_ written into UQL itself where there is no limit)

PR #194, opened 2023-11-08
[LS-53603] terraform support for service health panel

TheRealAkhil:
### Description
This PR adds support for the service health panel in the Lightstep terraform provider.

---
### Pre-merge checklist
- [x] add tests
- [x] verify behavior locally
- [x] update docs

---

R/CC


client/metric_dashboards.go
```diff
@@ -57,6 +58,14 @@ type Label struct {
 	Value string `json:"label_value"`
 }
 
+type Panel struct {
```
TheRealAkhil, line 61: this matches the publicapi implementation here

https://github.com/lightstep/lightstep/blob/1f3b9c5eabb01ab1061810fe95f6b86b969199ed/go/src/github.com/lightstep/crouton/service/api/public/v02/handlers/metric_dashboards.go#L150-L163

TheRealAkhil:
I'm sending this out for early review. I still need to add tests, update docs, verify behavior locally, etc.

Thanks for taking a look!

client/metric_dashboards.go
TheRealAkhil, line 61: this matches the publicapi implementation

lightstep/resource_metric_dashboard.go
```diff
@@ -890,6 +955,92 @@ func (p *resourceUnifiedDashboardImp) resourceUnifiedDashboardImport(ctx context
...
+			}
+			if sortDirection, ok := panelOptions["sort_direction"]; ok {
+				displayOptions["sort_direction"] = sortDirection
+			}
+			if percentile, ok := panelOptions["percentile"]; ok {
+				displayOptions["percentile"] = percentile
+			}
+			if changeSince, ok := panelOptions["change_since"]; ok {
+				displayOptions["change_since"] = changeSince
+			}
```
amburvill, line 987: Worth extracting this logic into a function that can be used across both `build` and `set` functions?

Something like:

```
func assignPanelOptions(from map[string]interface{}, to map[string]interface{}) {
	if sortBy, ok := from["sort_by"]; ok {
		to["sort_by"] = sortBy
	}
	if sortDirection, ok := from["sort_direction"]; ok {
		to["sort_direction"] = sortDirection
	}
	if percentile, ok := from["percentile"]; ok {
		to["percentile"] = percentile
	}
	if changeSince, ok := from["change_since"]; ok {
		to["change_since"] = changeSince
	}
}
````

File: lightstep/resource_metric_dashboard.go [hunk 3]
```diff
@@ -890,6 +955,92 @@ func (p *resourceUnifiedDashboardImp) resourceUnifiedDashboardImport(ctx context
...
+	}
+
+	return serviceHealthPanels, nil
+}
+
+func setServiceHealthPanelResourceData(
+	resource map[string]interface{}, // Terraform resource
+	panel client.Panel, // Panel from the API
+) {
+	resource["name"] = panel.Title
```
amburvill, line 1003: Maybe a comment here to signal this isn't a typo and we intentionally need to use `name` vs `title` in this case?

lightstep/resource_metric_dashboard.go
TheRealAkhil: Reading this function made me realize: we can probably just use the interface _as is_. i.e. skip the conversion entirely.

TheRealAkhil: Actually we may want to explicitly do this. We might add a new display type in the publicapi, and we won't want to return that automatically in the publicapi -> TF resource conversion.

amburvill (approved):

PR #197, opened 2023-12-19
docs: remove metric_condition, add composite alert examples

mattcarmody:
Removes examples from the docs for `metric_condition` resource because they're on the way out. Add an example for a composite alert to the `alert` resource.


docs/resources/alert.md
```diff
@@ -46,6 +46,67 @@ EOT
...
 ```
 
+## Example Usage (composite alert)
+
+```hcl
+resource "lightstep_alert" "high_error_rate_with_sustained_traffic" {
+  project_name = var.project
+  name         = "Error rates are high for my operation with sustained traffic"
+  description  = <<EOF
+Fires when over 20% of requests are erroring, but only when requests are above 75 spans over an hour window. This prevents firing when we have very low traffic and get a few error requests.
```
lambcode, line 56:
```suggestion
Fires when over 20% of requests have errors, but only when request count is above 75 over an hour window. This prevents firing when we have very low traffic and get a few error requests.
```

lambcode (approved):
Just a small verbiage suggestion :)

PR #199, opened 2024-01-03
alerts: add optional triggered for durations

mattcarmody:
Add support for optional triggered for durations: warning duration, critical duration and no data duration.

I didn't look into preventing it from being added to the metric condition resource, I assume it would be fairly messy because they share so much code.


lightstep/resource_alert.go
```diff
@@ -68,7 +68,7 @@ func getCompositeAlertFromUnifiedConditionResourceData(compositeAlertIn *client.
 		if subAlertIn.Expression.NoDataDurationMs != nil {
 			subAlertExpressionMap["no_data_duration_ms"] = subAlertIn.Expression.NoDataDurationMs
 		}
-		subAlerts = append(subAlerts, map[string]interface{}{})
+		subAlerts = append(subAlerts, subAlertExpressionMap)
```
lambcode, line 71: oops. Good catch

lambcode (approved):
Thanks for fixing all the bugs!

PR #201, opened 2024-01-11
fix hidden_queries plan bug

matthagenbuch:
Fixes a bug when using the `lightstep_alert.query.hidden_queries` attribute. Currently, if this attribute is set the plan will always show a change because it isn't being read during the refresh stage. This change reads the hidden_queries into the resource during refresh, resulting in a clean plan.


MisterSquishy (approved):
weirdddd how was [this test](https://github.com/lightstep/terraform-provider-lightstep/blob/7ef0a8e978164b4353f842f9dece0d3b943bdac4/lightstep/resource_dashboard_test.go#L658-L659) working without a `ExpectNonEmptyPlan`? did we add handling for this in dashboards but forget about alerts? 

either way maybe worth an analogous test in alerts!

MisterSquishy:
oh also dont forget to update `.go-version`!

PR #206, opened 2024-02-06
deprecate environment in favor of api_url

matthagenbuch:
This is meant to add support for provisioning resources in the new EU instances. Currently, the provider accepts an `environment` parameter which must be public|meta|staging , and we construct the API URL based on this. 

To make this change, I deprecated the environment parameter and added an `api_url` parameter which always takes precedence. If neither is provided, it defaults to the US public instance (as it does today). Terraform will display a deprecation warning for those using the provider with the environment parameter, and we would eventually remove it.


client/client.go
```diff
@@ -74,25 +74,11 @@ type Client struct {
 }
 
 // NewClient gets a client for the public API
-func NewClient(apiKey string, orgName string, env string) *Client {
-	return NewClientWithUserAgent(apiKey, orgName, env, fmt.Sprintf("%s/%s", DefaultUserAgent, version.ProviderVersion))
+func NewClient(apiKey string, orgName string, url string) *Client {
```
brettbuddin, line 77: 🎨 `baseURL` for consistency here, too?

exporter/exporter.go
```diff
@@ -169,9 +169,9 @@ func Run(args ...string) error {
 	}
 
 	// default to public API environment
-	lightstepEnv := "public"
-	if len(os.Getenv("LIGHTSTEP_ENV")) > 0 {
-		lightstepEnv = os.Getenv("LIGHTSTEP_ENV")
+	lightstepUrl := "https://api.lightstep.com"
```
brettbuddin, line 172: 🎨 Hoist this up as a constant

lightstep/provider.go
```diff
@@ -30,7 +30,14 @@ func Provider() *schema.Provider {
...
 				DefaultFunc:  schema.EnvDefaultFunc("LIGHTSTEP_ENV", "public"),
 				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^(staging|meta|public)$`), "Must be one of: staging, meta, public"),
-				Description:  "The name of the Lightstep environment, must be one of: staging, meta, public.",
+				Description:  "(deprecated) The name of the Lightstep environment, must be one of: staging, meta, public. Deprecated in favor of `api_url`",
+				Deprecated:   "This field is deprecated and will be removed in a future release. Please use the `api_url` field instead.",
+			},
+			"api_url": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("LIGHTSTEP_API_BASE_URL", ""),
```
brettbuddin, line 39: Thoughts on whether this should be called `api_base_url` or if the environment variable should be named `LIGHTSTEP_API_URL` to be consistent?

lightstep/provider.go
```diff
@@ -93,10 +100,24 @@ func configureProvider(_ context.Context, d *schema.ResourceData) (interface{},
 		apiKey = apiKeyEnv
 	}
 
+	var baseUrl string
```
brettbuddin, line 103: 🎨 https://google.github.io/styleguide/go/decisions.html#initialisms

matthagenbuch:
@brettbuddin stil very much a work in progress but thanks for the review! i'm currently struggling with the acceptance tests / doc generation 😕

lightstep/provider.go
matthagenbuch: It probably should, but I wanted to reuse the current environment variable to avoid breaking changes

exporter/exporter.go
matthagenbuch: I brought this logic in from the client.go file and left a comment to remove it once `environment` is deprecated. I'd rather leave the URL in-line to make it easier to clean up later. WDYT?

.github/workflows/pr.yml
```diff
@@ -54,6 +54,7 @@ jobs:
         run: |
           rm -r docs
           /tmp/tfplugindocs
+          git diff
```
matthagenbuch, line 57: this helps identify the doc diff for those of us who have trouble running it locally

exporter/exporter.go
brettbuddin: Yeah, the only concern would be that this URL is mentioned in multiple places and things diverge. If the risk is low there, feel free to ignore.

lightstep/provider.go
brettbuddin: Ah okay. I don't think that's necessarily a problem if it's existed for a while. The only question I'd have is: Are we changing the meaning of the original environment variable? I don't think we are, but figured I'd ask.

brettbuddin: Perhaps we could have it support both `LIGHTSTEP_API_URL` and `LIGHTSTEP_API_BASE_URL`? If one isn't set, then look for the other. People coming in fresh would then not have to care about the old way.

matthagenbuch: good point, looks like there's a helper function for this. updated.

exporter/exporter.go
```diff
@@ -169,9 +169,9 @@ func Run(args ...string) error {
...
+	if len(os.Getenv("LIGHTSTEP_API_BASE_URL")) > 0 {
```
brettbuddin, line 173: I suppose we'll need to alter this to support `LIGHTSTEP_API_URL`, too

matthagenbuch: ah good catch

client/client_test.go
```diff
@@ -1,26 +1,27 @@
...
 
 func TestNew_other(t *testing.T) {
 	t.Parallel()
-	c := NewClient("api-key", "org-name", "other")
-	assert.Equal(t, "https://api-other.lightstep.com/public/v0.2/org-name", c.baseURL)
+	c := NewClient("api-key", "org-name", "https://api-other.lightstep.com")
+	assert.Equal(t, "https://api-other.lightstep.com/public/v0.2/org-name", c.baseUrl)
 }
 
 func TestNew_env_var_provided_baseURL(t *testing.T) {
```
brettbuddin, line 22: This test is no longer testing the same thing. Delete it?

exporter/exporter.go
```diff
@@ -169,9 +169,11 @@ func Run(args ...string) error {
...
 	// default to public API environment
-	lightstepEnv := "public"
-	if len(os.Getenv("LIGHTSTEP_ENV")) > 0 {
-		lightstepEnv = os.Getenv("LIGHTSTEP_ENV")
+	lightstepUrl := "https://api.lightstep.com"
+	if len(os.Getenv("LIGHTSTEP_API_URL")) > 0 {
+		lightstepUrl = os.Getenv("LIGHTSTEP_API_URL")
+	} else if len(os.Getenv("LIGHTSTEP_API_BASE_URL")) > 0 {
+		lightstepUrl = os.Getenv("LIGHTSTEP_API_BASE_URL")
 	}
```
brettbuddin, line 177: I just reviewed the tests and this appears to be untested now that we've moved it out of the client. Might need to hoist some of this environment handling out into a helper that you can unit test—unless there's a clear path to verify the configuration combinatorics in this `Run` function directly.

brettbuddin: Yeah, I agree testing `Run` is out of scope. I could make a case that since this aspect was somewhat under test before that we've lost some coverage, but it's relatively constrained and we can address in another pass if you'd like.

matthagenbuch: Maybe i was just being lazy, I added a test for Run

exporter/exporter_test.go
```diff
@@ -100,3 +105,30 @@ EOT`,
...
+		_, err = w.Write(body)
+		assert.NoError(t, err)
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer server.Close()
+
+	os.Setenv("LIGHTSTEP_API_KEY", "api-key")
+	os.Setenv("LIGHTSTEP_ORG", "blars-org")
+	os.Setenv("LIGHTSTEP_API_URL", server.URL)
+	Run("exporter", "export", "lightstep_dashboard", "blars", "123abc")
```
brettbuddin, line 133: Thanks for adding this! Just two things I would add here:

1. An assertion for `nil` error return value from `Run`.
2. Communication between the harness and the HTTP handler so you know its your handler that's exercised and not some other handler in the event of a healthy result from `Run`. This is a synchronous action so you could use an atomic for that, but you could use something like this, too:

```go
done := make(chan struct{})

server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    defer close(done)
    // Remainder of HTTP handler...
}))

// All environment setting and Run call

select {
case <-time.After(10 * time.Second):
    t.Fatal("timeout waiting for HTTP handler to be exercised")
case <-done:
}

```

brettbuddin (approved):

PR #208, opened 2024-02-08
Increase Big Number subtitle maximum length

bcronin:
Updates the Big Number subtitle maximum length to correspond to updates to Cloud Observability.

_Note: a new release has not been created with a corresponding version increment for this change.  This small update is intended to go out with the next release._


MisterSquishy (approved):
Big Numbers need Big Subtitles

PR #209, opened 2024-02-12
client: support EU environments

gdvalle:
Add new env strings for `LIGHTSTEP_ENV` to support the EU instances `eu-public` and `eu-staging`. Add a unit test for the two new envs.


gdvalle:
Oh, now I see https://github.com/lightstep/terraform-provider-lightstep/pull/206...closing in favor of that.

PR #211, opened 2024-02-21
Fix bug where warning threshold duration could not be applied

lambcode:
One of the checks for the warning threshold was checking the critical threshold duration rather than the warning threshold duration. This made it impossible to specify just one of the durations as setting just error duration would cause a nil pointer dereference and setting just warning duration would be ignored.


mattcarmody (approved):
LGTM thank you!

PR #213, opened 2024-02-22
fix go-version staling

MisterSquishy:
adds a CI check to make sure the version is incremented

CleanShot 2024-02-22 at 10 41 55@2x


.go-version
```diff
@@ -1,2 +1,2 @@
-1.91.2
```
MisterSquishy, line 1: i merged with `1.91.2`, which turned out to be taken, so i lazily published it as `1.91.3`. thus the jump to .4

paivagustavo (approved):

PR #217, opened 2024-03-11
Add display_type to display_type_options

erik-snow:
The frontend adds `display_type` to `display_type_options`. If a customer attempts to import a chart into Terraform, it will fail because that key will exist in the chart but it's not supported in TF.


MisterSquishy (approved):
🚢

bcronin (approved):

PR #218, opened 2024-03-26
Add snooze rules resource

lambcode:
Adds snooze rules as a resource to the terraform provider. 

Note that I had to fix how we generate docs now that we have tofu as our terraform binary. The docs can still be created via the `make docs` command, and should now be agnostic to whether you have terraform or tofu installed.

@lightstep/team-alerts-be


DEVELOPMENT.md
```diff
@@ -38,6 +38,6 @@ Documentation files in `docs` are generated by combining:
 Any edits to documentation files should be made in the `templates` directory or the .go files. Then you can use the following steps to generate the files in the docs directory:
 
 1. Download `tfplugindocs` from https://github.com/hashicorp/terraform-plugin-docs/releases and make it available on your `PATH`
-2. Run `$ make docs` or `$ tfplugindocs` from the root of this directory
```
lambcode, line 41: Running tfplugindocs directly doesn't work anymore because tfplugindocs builds the provider with the `registry.terraform.io` prefix, but then calls the terraform binary which is now open-tofu that looks for plugins with the prefix of `registry.opentofu.org`.

Makefile
```diff
@@ -30,8 +29,9 @@ fmt:
 	go fmt
 
 .PHONY: docs
-docs:
-	tfplugindocs
+docs: build
+	$(MAKE) -C tfplugindocs_working generate-schema
+	tfplugindocs generate --providers-schema tfplugindocs_working/schema.json
```
lambcode, line 34: Now we generate the schema separately from generating the docs to allow us bypass the `tfplugindocs` auto compiling the provider.

gontzess (approved):
LGTM!

PR #219, opened 2024-04-02
alerts: remove tests using matchon filter

mattcarmody:
Remove use of the matchon feature for alerting rules from the metric condition tests. This feature is no longer documented/supported.


gontzess (approved):
🚀

PR #221, opened 2024-04-08
alerts: remove matchon support from alerting rules

mattcarmody:
The feature is unused, it was never adopted by anyone other than ourselves and it is not supported in the UI.


lambcode (approved):
It's happening!

PR #222, opened 2024-05-03
Fix spurious diff with `hidden_queries`

MisterSquishy:
Fixes a spurious diff issue, in which alerts with `hidden_queries` set will continuously try (and fail) to "update". This is caused by a failure to store the `hidden_queries` map in state, causing an unresolvable diff between the HCL and the state file.

I fixed this by stealing the `hidden_queries` setting code from `lightstep_dashboard` and calling it from `lightstep_alert`, and added a unit test to ensure the plan is clean for these types of alerts now


lightstep/util.go
```diff
@@ -21,3 +25,21 @@ func convertNestedMapToSchemaSet(opts map[string]interface{}) *schema.Set {
...
+// Note due to Terraform's issues with TypeMap having TypeBool elements, we
+// need to use boolean strings
+func setHiddenQueriesFromResourceData(qs map[string]interface{}, query client.MetricQueryWithAttributes) {
+	if len(query.HiddenQueries) == 0 {
+		// nothing to do
+		return
+	}
+	hq := make(map[string]interface{}, len(query.HiddenQueries))
+	for k, v := range query.HiddenQueries {
+		// Don't include the top-level query in the TF resource data
```
MisterSquishy, line 38: note that this leads to another unresolvable diff scenario, specifically when the user specifies a `hidden_queries` with the top-level query name included. i am choosing not to fix this in the provider, because ~i dont know how~ i think the API should be updated to better handle this case -- in particular, we should omit the top-level query name from `hidden_queries` in the API, and use `hidden` for that purpose (as the contract has been defined here)

danhurwit (approved):

PR #227, opened 2024-05-31
Add Tricharts

erik-snow:
Adds `display_type: tricharts`


lightstep/resource_dashboard_test.go
```diff
@@ -933,6 +933,53 @@ chart {
...
+chart {
+	name   = "Chart #123"
+	type   = "timeseries"
+	rank   = 0
+	x_pos  = 4
+	y_pos  = 0
+	width  = 4
+	height = 4
+
+	query {
```
MisterSquishy, line 956: is there any validation that the user actually added 3 queries, or does the FE degrade gracefully when they fail to do so? id be happy with either/both but am less thrilled about neither

MisterSquishy (approved):
🚢

lightstep/resource_dashboard_test.go
erik-snow: The FE will show an invalid query error if there are not 3 queries and give an example of a valid query. Not ideal but consistent with other query/viz pairings.

PR #231, opened 2024-06-24
LS-61620: add dashbaord chart thresholds to provider

srjames90:
Adds chart thresholds to the provider


lightstep/resource_metric_dashboard.go
```diff
@@ -599,6 +611,78 @@ func buildCharts(chartsIn []interface{}) ([]client.UnifiedChart, error) {
...
+	"#6699CC": true,
+}
+
+var operators = map[string]bool{
+	"E":  true,
+	"GE": true,
+	"GT": true,
+	"LE": true,
+	"LT": true,
+}
```
srjames90, line 636: where do consts usually live?

.go-version
```diff
@@ -1 +1 @@
-1.95.8
+1.95.9
```
MisterSquishy, line 1: nit: additive changes get a minor bump
```suggestion
1.96.0
```

lightstep/resource_dashboard.go
```diff
@@ -9,6 +9,49 @@ import (
...
+				"#DC7847",
+				"#78469B",
+				"#37A2AE",
+				"#B03B7F",
+				"#1F40C1",
+				"#8B7255",
+				"#826CEF",
+				"#D56DD5",
+				"#6699CC",
+			}, false),
```
MisterSquishy, line 31: wait i thought we were going to allow people to pass anything they wanted here?
```suggestion
```

lightstep/resource_metric_dashboard.go
MisterSquishy: i think the `ValidateFunc` above makes this const unnecessary, right?

erik-snow: Could/should the colors live in backend validation instead of here? If we add more colors/operators, TF would need to be updated as well.

lightstep/resource_metric_dashboard.go
```diff
@@ -599,6 +611,78 @@ func buildCharts(chartsIn []interface{}) ([]client.UnifiedChart, error) {
...
+		threshold := t.(map[string]interface{})
+		color, ok := threshold["color"].(string)
+
+		if !ok {
+			return []client.Threshold{}, fmt.Errorf("missing required attribute 'color' for threshold")
+		}
+
+		if !colors[color] {
+			return []client.Threshold{}, fmt.Errorf("invalid value for attribute 'color' for threshold")
+		}
```
MisterSquishy, line 653: i dont think we should validate this one, but also i think the `ValidateFunc` will do this for you
```suggestion
```

lightstep/resource_metric_dashboard.go
```diff
@@ -599,6 +611,78 @@ func buildCharts(chartsIn []interface{}) ([]client.UnifiedChart, error) {
...
+
+		operator, ok := threshold["operator"].(string)
+		if !ok {
+			return []client.Threshold{}, fmt.Errorf("missing required attribute 'operator' for threshold")
+		}
+
+		if !operators[operator] {
+			return []client.Threshold{}, fmt.Errorf("invalid value for attribute 'operator' for threshold")
+		}
```
MisterSquishy, line 662: ditto here re: `ValidateFunc`
```suggestion
```

lightstep/resource_metric_dashboard.go
```diff
@@ -599,6 +611,78 @@ func buildCharts(chartsIn []interface{}) ([]client.UnifiedChart, error) {
...
+
+		label, ok := threshold["label"].(string)
+
+		if !ok {
+			return []client.Threshold{}, fmt.Errorf("missing required attribute 'label' for threshold")
```
MisterSquishy, line 667: i thought this one was optional?

MisterSquishy: 🤔 how is the unit test passing when it omits this? is `nil.(string)` just `""`?

lightstep/resource_metric_dashboard.go
srjames90: I can remove the validation here and move it to the other PR I have open where I'm doing the updates BE updates for updating

MisterSquishy (commented):
🔥 just a few nitpicks!

lightstep/resource_dashboard.go
srjames90: Currently, it's a dropdown of options, but I think there was talk of adding a color wheel at some point, so I can remove this

srjames90: Yup, it's currently limited

lightstep/resource_metric_dashboard.go
srjames90: taking a look, but yeah, I think it gets converted to an empty string

lightstep/resource_metric_dashboard.go
MisterSquishy: i think _if_ we plan on validating colors, it should be here -- that way we can catch invalid colors at plan time instead of waiting until you try to apply. that said, i still think restricting the list of colors is a much worse UX for little benefit

lightstep/resource_dashboard.go
MisterSquishy: we _want_ to limit the allowed values in terraform? why? is the concern just that we wouldn't know how to show a custom color in the chart config item's list? that seems like a minor concern compared to not being able to configure the color i want

srjames90: removed

lightstep/resource_metric_dashboard.go
MisterSquishy: 🤷 yeah that seems fine by me. we are historically super bad about distinguishing between nil vs zero (i.e. empty string). id recommend just typing `Label` as a `string` instead of a pointer to a string, since we seem to never use `nil`

client/metric_dashboards.go
```diff
@@ -67,6 +68,15 @@ type Panel struct {
 	Body     map[string]any  `json:"body"`
 }
 
+type Threshold struct {
+	// enum: E,GE,GT,LE,LT
+	Operator string  `json:"operator"`
+	Value    float64 `json:"value"`
+	// An alpha hex color
+	Color string  `json:"color"`
+	Label *string `json:"label"`
```
MisterSquishy, line 77: re: https://github.com/lightstep/terraform-provider-lightstep/pull/231/files#r1651587958
```suggestion
	Label string `json:"label"`
```

lightstep/resource_metric_dashboard.go
MisterSquishy: (and then we can remove this short-circuit since it never actually gets hit)
```suggestion
```

MisterSquishy (approved):

